PR #200, opened 2026-04-30
fix: [*MouseState.UpdateFrame] zeroes scrollX and scrollY when before calling onUpdate handler

### LINKED ISSUE #199: Bug?: Mouse scroll value is zero in `app.OnUpdate` handler
I got a problem, MouseState scrollX, scrollY is zero when I do `app.Input().Mouse().Scroll()` in `app.OnUpdate` handler.

Image

Because I think `MouseState#UpdateFrame` seem sets scrollX and scrollY to zero when just before `onUpdate` in `app.go:306`.

https://github.com/gogpu/gogpu/blob/cd55b8da906e471ca3d08c23d8e60b0dc0ecf692/input/mouse.go#L138-L146

https://github.com/gogpu/gogpu/blob/cd55b8da906e471ca3d08c23d8e60b0dc0ecf692/input/input.go#L25-L29

https://github.com/gogpu/gogpu/blob/cd55b8da906e471ca3d08c23d8e60b0dc0ecf692/app.go#L302-L313

I've an idea for fix this problem, I think `MouseState#UpdateFrame` to be don't zeroes scrollX and scrollY, When mouse scrolled in current frame.

And I'll create Pull Request to fix this bug for your awesome project.

(my english isn't good, and my english contains some rude expressions probably. sorry)

----

My go version: `go version go1.26.2 darwin/arm64`.

Minimum example to reproducing the problem.
```go
package main

import (
	"fmt"

	"github.com/gogpu/gogpu"
	"github.com/gogpu/gogpu/gmath"
)

func main() {
	app := gogpu.NewApp(gogpu.DefaultConfig().
		WithTitle("Hello GoGPU").
		WithSize(800, 600))

	col := gmath.DarkGray

	app.OnDraw(func(dc *gogpu.Context) {
		dc.DrawTriangleColor(col)
	})

	app.OnUpdate(func(dt float64) {
		mouse := app.Input().Mouse()

		wx, wy := mouse.Scroll()
		fmt.Printf("wx=%f, wy=%f\n", wx, wy)

		// Set background color to red when mouse scrolled, But it never change.
		if wx != 0 || wy != 0 {
			col = gmath.Red
		}
	})

	app.Run()
}
```

k-chimi:
resolve #199

### DIFF
--- a/input/mouse.go
+++ b/input/mouse.go
@@ -22,6 +22,7 @@ type MouseState struct {
 	mu               sync.RWMutex
 	x, y             float32
 	prevX, prevY     float32
+	isScrolled       bool
 	scrollX, scrollY float32
 	current          [MouseButtonCount]bool
 	previous         [MouseButtonCount]bool
@@ -55,6 +56,8 @@ func (m *MouseState) SetButton(button MouseButton, pressed bool) {
 func (m *MouseState) SetScroll(x, y float32) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+
+	m.isScrolled = true
 	m.scrollX = x
 	m.scrollY = y
 }
@@ -141,6 +144,10 @@ func (m *MouseState) UpdateFrame() {
 	m.previous = m.current
 	m.prevX = m.x
 	m.prevY = m.y
-	m.scrollX = 0
-	m.scrollY = 0
+
+	if !m.isScrolled {
+		m.scrollX, m.scrollY = 0, 0
+	}
+
+	m.isScrolled = false
 }
--- a/input/mouse.go
+++ b/input/mouse.go
@@ -19,13 +19,13 @@ const (
 // MouseState holds mouse input state.
 // All methods are thread-safe.
 type MouseState struct {
-	mu               sync.RWMutex
-	x, y             float32
-	prevX, prevY     float32
-	isScrolled       bool
-	scrollX, scrollY float32
-	current          [MouseButtonCount]bool
-	previous         [MouseButtonCount]bool
+	mu                         sync.RWMutex
+	x, y                       float32
+	prevX, prevY               float32
+	scrollX, scrollY           float32
+	frameScrollX, frameScrollY float32
+	current                    [MouseButtonCount]bool
+	previous                   [MouseButtonCount]bool
 }
 
 func newMouseState() MouseState {
@@ -57,9 +57,8 @@ func (m *MouseState) SetScroll(x, y float32) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	m.isScrolled = true
-	m.scrollX = x
-	m.scrollY = y
+	m.scrollX += x
+	m.scrollY += y
 }
 
 // Position returns current mouse position.
@@ -99,7 +98,8 @@ func (m *MouseState) Delta() (dx, dy float32) {
 func (m *MouseState) Scroll() (x, y float32) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
-	return m.scrollX, m.scrollY
+
+	return m.frameScrollX, m.frameScrollY
 }
 
 // Pressed returns true if button is currently pressed.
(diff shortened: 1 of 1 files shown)


kolkov (changes requested):
Hi @k-chimi, thanks for the excellent bug report in #199 and the PR! You correctly identified the root cause — `UpdateFrame()` zeroes scroll before `OnUpdate` reads it.

### The bug is real, but the fix needs a different approach

The `isScrolled` flag solves the immediate symptom, but has two issues:

1. **No accumulation.** `SetScroll` still overwrites (`m.scrollX = x`). If two scroll events arrive in the same frame (common on macOS trackpads with high-frequency scroll), the first is lost. It should be `m.scrollX += x`.

2. **Per-field flags don't scale.** When we add touch/gesture input, we'd need `isTouched`, `isPinched`, etc. The real problem is architectural — the reset/read ordering.

### Architecture context

gogpu provides **dual input models**:

- **Event-driven** (for GUI apps): `SetScrollCallback` delivers each event immediately — this path works correctly, no bug here.
- **Polled** (for games, Ebiten-style): `Mouse().Scroll()` in `OnUpdate` reads a per-frame snapshot — this is where the bug is.

The current frame order:

```
processEvents()     → platform calls SetScroll() (possibly multiple times)
inputState.Update() → zeroes scrollX/scrollY  ← bug: data lost
onUpdate(dt)        → user calls Scroll() → gets (0, 0)
```

### Enterprise reference research

We checked 6 frameworks to find the correct pattern for the polled model:

| Framework | Input Model | Accumulates scroll? | Double-buffer? |
|-----------|-------------|---------------------|----------------|
| SDL3 | Event queue | No | No |
| winit (Rust) | Event dispatch | No | No |
| Qt6 | QWheelEvent | No | No |
| GTK4 | Signal dispatch | No | No |
| GLFW | Callback | No | No |
| **Ebiten** | **Polled game tick** | **Yes (`+=`)** | **Yes (`copyAndReset`)** |

5 of 6 are event-driven and don't need accumulation — the app handles each event individually. Only **Ebiten** accumulates, because it uses a polled game-tick model where the app reads a snapshot once per frame. Our polled path (`OnUpdate` + `Mouse().Scroll()`) is the same model, so we need the same pattern.

### Correct fix (Ebiten's copyAndReset pattern)

Two changes:

**1. Accumulate in SetScroll:**
```go
func (m *MouseState) SetScroll(x, y float32) {
    m.mu.Lock()
    defer m.mu.Unlock()
    m.scrollX += x  // accumulate all events within a frame
    m.scrollY += y
}
```

**2. Snapshot + reset in UpdateFrame:**
```go
type MouseState struct {
    // ...
    scrollX, scrollY         float32 // accumulator (written by platform callbacks)
    frameScrollX, frameScrollY float32 // snapshot (read by user in OnUpdate)
}

func (m *MouseState) UpdateFrame() {
    m.mu.Lock()
    defer m.mu.Unlock()
    m.previous = m.current
    m.prevX = m.x
    m.prevY = m.y
    m.frameScrollX = m.scrollX  // snapshot accumulated deltas
    m.frameScrollY = m.scrollY
    m.scrollX = 0               // reset accumulator for next frame
    m.scrollY = 0
}

func (m *MouseState) Scroll() (x, y float32) {
    m.mu.RLock()
    defer m.mu.RUnlock()
    return m.frameScrollX, m.frameScrollY  // read snapshot, not accumulator
}
```

This gives us the correct frame order:
```
processEvents()     → SetScroll accumulates (+= each event)
inputState.Update() → snapshot: frameScroll = scroll; scroll = 0
onUpdate(dt)        → Scroll() returns frameScroll ✓
```

No `isScrolled` flag needed, handles multiple events per frame, and follows the Ebiten pattern exactly.

### Next steps

Would you like to update this PR with the accumulate + snapshot approach? The changes are contained to `input/mouse.go` — just the three methods above (`SetScroll`, `UpdateFrame`, `Scroll`) plus adding the two `frameScroll` fields to the struct. We're happy to help if you have any questions along the way.

Your bug report and analysis in #199 were spot-on — the code references to `app.go:306`, `mouse.go:138-146`, and `input.go:25-29` made the root cause immediately clear. Great first contribution, and welcome to the project!

k-chimi:
Thanks! I'il fix my pr for follow your collect inspection!

k-chimi:
\('v')/ I've fix my code, Please check it when you have a moment. 🙏
image

kolkov (approved):
Accumulate + snapshot pattern implemented correctly, exactly as suggested. Clean diff, no unnecessary changes. LGTM!

Issue #203, opened 2026-05-04
.Xauthority and $DISPLAY parsing mismatch?


sverrehu:
Hi! I wonder if there is a mismatch between parsing of addresses in [the .Xauthority parser](https://github.com/gogpu/gogpu/blob/9bad8d547bcf84131af956fdfbb0b6e943a5abfa/internal/platform/x11/auth.go#L118) and the [DISPLAY environment parser](https://github.com/gogpu/gogpu/blob/9bad8d547bcf84131af956fdfbb0b6e943a5abfa/internal/platform/x11/connection.go#L125).

It appears the former will result in a string in which each rune has the value of an octet from an IP address, while the latter will contain the ASCII representation of the octets of an address, including dots.

I believe the parsing of the address in `.Xauthority` may be wrong. Here is an example file containing an IPv4 address:

```text
$ touch sample.Xauthority
$ xauth -f sample.Xauthority add 192.168.0.1:0 . $(mcookie)
$ od -tx1 sample.Xauthority 
0000000 00 00 00 04 c0 a8 00 01 00 01 30 00 12 4d 49 54
0000020 2d 4d 41 47 49 43 2d 43 4f 4f 4b 49 45 2d 31 00
0000040 10 74 a6 67 bb 67 7d 5e 08 0b 6f ee 36 c4 c2 ef
0000060 cf
0000061
```

As can be seen, the `192.168.0.1` address is represented by the four bytes `c0 a8 00 01`, which will end up as four runes in the address representation.

When the code later does [a string-by-string comparison](https://github.com/gogpu/gogpu/blob/9bad8d547bcf84131af956fdfbb0b6e943a5abfa/internal/platform/x11/auth.go#L238) of the two values, they will not match even if they should.

I believe this may be the reason I get _"x11: connection setup failed: Authorization required, but no authorization protocol specified"_ when I run gogpu code on a Linux machine that connects to an X11 server on a different IP address (`DISPLAY=192.168.0.1:0`).

Disclaimer: I'm still new to Go, so I might read your code incorrectly.

kolkov:
Thanks, I'll have to do some research. I'll get to it as soon as I have some free time.

kolkov:
Thanks @sverrehu, great catch! You were absolutely right.

[...]

PR #204, opened 2026-05-04
fix(x11): binary .Xauthority address matching for remote displays

kolkov:
## Summary

- **Fix remote X11 display authentication** (#203, @sverrehu) — `.Xauthority` stores IPv4 as raw 4 bytes, but DISPLAY parser produced ASCII strings. Fixed with libxcb `getpeername()` pattern. Added IPv6 support. First Pure Go X11 library to handle `FamilyInternet` correctly.
- **Fix all goconst/gocritic lint issues** — 0 lint issues on all 3 platforms (Linux/Windows/Darwin).

## Test plan

- [x] `GOOS=linux go build ./...` — compiles
- [x] `GOOS=linux go vet ./...` — clean
- [x] `golangci-lint run` — 0 issues on linux/windows/darwin
- [ ] CI green on all platforms
- [ ] @sverrehu: test with remote `DISPLAY=<ip>:0` setup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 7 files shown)


sverrehu:
Thank you! Will test on my Linux machine when I'm back at work tomorrow morning.

Issue #213, opened 2026-05-08
OnWindowClose callback behavior and special cases


lkmavi:
@kolkov And hello again, I added something

##**Behavior and threading**
**Registration**

- OnWindowClose(fn func(*Window)) stores the callback on the adapter.

Dispatch semantics

- dispatchWindowClose(w *Window) invokes the registered callback if non‑nil; otherwise it is a no‑op.

- The callback is invoked on the event loop / render thread where platform close events are processed. It is invoked before platform resources are destroyed so the callback can safely read window metadata.

Threading guidance

- Callbacks run on the event/render thread. They must be fast and non‑blocking.

- If heavy work is required, spawn a goroutine from the callback and synchronize access to shared state.

## Special case: recalculating window IDs and creating a replacement window
When this happens

- You need to recalculate or compact window IDs (for example, to keep a contiguous index set) or open a replacement window using the closed window’s saved settings (size, position, layout, user preferences).

**Recommended pattern**

1. Read state synchronously inside the OnWindowClose callback (safe because dispatch happens before teardown). Read the window’s ID, persisted settings, and any bookkeeping keys.

2. Defer heavy work: if recalculation or persistence is expensive, spawn a goroutine to perform it. Use channels or mutexes to coordinate with the main data structures.

3. Perform ID recomputation in a single place (e.g., windowManager.reindex()), and call that from the goroutine or from the event loop after the callback returns. Avoid doing ad‑hoc ID changes from multiple callbacks concurrently.

4. Create replacement windows on the event loop: if you must open a new window with the closed window’s settings, schedule that creation on the event loop (e.g., push a task into the app’s event queue or call app.NewWindow from the event thread). This ensures platform creation happens on the correct thread and avoids races with teardown.

5. Order of operations: prefer this sequence:

- read and snapshot closed window metadata inside OnWindowClose,

- return quickly from the callback,

- perform recompute/create work on the event loop or a synchronized goroutine,

- call platform CreateWindow only from the event loop (or from a goroutine that dispatches to the main thread).

lkmavi:
Hi @kolkov I'm back after studying the discussion and the current code.
I’m ready to start working on it – just let me know if there are any constraints or preferences you have in mind.

## Tabbing Project
Currently creating new tabs via app.NewWindow(tabConfig(...)) works perfectly with the native tabbing system.
Missing capabilities:

Intercepting the close of a specific tab (to save state or return its ID to a pool).

[...]

Issue #210, opened 2026-05-07
Per‑window input callbacks for multi‑window applications


lkmavi:
Hello again, and thanks once more for the fantastic work on gogpu — the multi‑window rendering and tabbing support already feels very solid, and I’m really enjoying building on it.

As a follow‑up to our earlier discussion about my project (the tabbed application), I wanted to open a dedicated feature request regarding input events for secondary windows, because I might have missed something obvious. If there’s already a way to achieve this, I’d be grateful if you could point me to it.

The situation

I’m creating multiple windows with NewWindow(), all sharing the same TabbingIdentifier. The primary window’s event source works perfectly, but once focus moves to a secondary window (by clicking on it), the global keyboard shortcuts (like Cmd+N / Ctrl+N) stop firing, because the key callbacks are only wired on the primary window inside App.setupInputEvents().

**Minimal reproduction (current behavior)**
```go
func main() {
    app := gogpu.NewApp(
        gogpu.DefaultConfig().WithTitle("Main").
            WithTabbingMode(gogpu.TabbingPreferred).
            WithTabbingIdentifier("com.demo"),
    )
    var created bool
    events := app.EventSource()
    events.OnKeyPress(func(key gpucontext.Key, mods gpucontext.Modifiers) {
        if key == gpucontext.KeyN && (mods&gpucontext.ModControl != 0) {
            if !created {
                created = true
                app.NewWindow(gogpu.DefaultConfig().
                    WithTitle("Second").
                    WithTabbingMode(gogpu.TabbingPreferred).
                    WithTabbingIdentifier("com.demo"))
            } else {
                // This never prints after focus moves to the second window
                println("third window requested")
            }
        }
    })
    app.Run()
}
```
After clicking on the second window, the primary window loses focus, and Cmd+N no longer reaches the event source — so no further windows can be created.

What I found in the source

App.setupInputEvents() (app.go) sets SetKeyCallback (and pointer/scroll/char) only on a.platWindow.

App.NewWindow() creates a new platform.PlatformWindow but does not copy or assign any input callbacks.

The Window struct has SetOnDraw, SetOnResize, SetOnClose, but no way to subscribe to input events.

The PlatformWindow interface already includes SetKeyCallback, SetPointerCallback, etc., so the underlying machinery is all there.

1. Desired API (any of the following would be perfect)
```go
w2, _ := app.NewWindow(...)
w2.OnKeyPress(func(key gpucontext.Key, mods gpucontext.Modifiers) {
[...]

PR #214, opened 2026-05-08
fix/213 : Add OnWindowClose callback

lkmavi:


kolkov (commented):
Hi @lkmavi, thank you for identifying this gap — you're absolutely right that window close events need to be dispatched to application code before resources are destroyed. The dispatch-before-teardown timing and the tests are great.

**Small ask:** could you add a PR description with a summary and link to #213? It helps with changelog generation and makes the PR history easier to navigate.

After researching how winit, SDL3, Qt6, GTK4, Wails 3, and GLFW handle this, I think we should discuss the architecture before merging. Here's what we found:

## The core issue

`Window.SetOnClose(func() bool)` already exists in `window_manager.go` — it's the correct per-window close callback with rejection support (return `false` to cancel close). But **`classifyEvent` never calls it**. That's the actual bug.

Every enterprise framework puts close callbacks on the per-window object, not on an app-level event source:

| Framework | Close callback location | Can reject? |
|-----------|------------------------|-------------|
| winit | `WindowEvent::CloseRequested` per window | Yes (ignore event) |
| SDL3 | `SDL_EVENT_WINDOW_CLOSE_REQUESTED` per window | Yes (ignore event) |
| Qt6 | `QWindow::closeEvent()` virtual method | Yes (`event->ignore()`) |
| GTK4 | `close-request` signal per GtkWindow | Yes (return TRUE) |
| Wails 3 | `RegisterHook(WindowClosing)` per window | Yes (`event.Cancel()`) |
| GLFW | `glfwSetWindowCloseCallback(window, fn)` | Yes (clear flag) |

## What I think we should do instead

**Step 1:** Wire the existing `Window.SetOnClose` into `classifyEvent`:

```go
case platform.EventClose:
    if isPrimary {
        if a.primaryWindow != nil && a.primaryWindow.onClose != nil {
            if !a.primaryWindow.onClose() {
                break // close rejected
            }
        }
        a.running = false
    } else {
        w := a.windowManager.get(event.WindowID)
        if w != nil && w.onClose != nil {
            if !w.onClose() {
                break // close rejected
            }
        }
        a.closeSecondaryWindow(event.WindowID)
    }
```

This gives you exactly what you need for your tabbing use case:
```go
w2, _ := app.NewWindow(config)
w2.SetOnClose(func() bool {
    // persist tab state, return tab ID to pool, etc.
    saveTabState(w2.ID())
    return true // allow close
})
```

**Step 2 (future):** For cross-window awareness through `gpucontext.EventSource` (needed by UI toolkit), we'd add a WindowID-based callback — but that's a separate API design discussion.

## Why not `OnWindowClose(func(*Window))` on EventSource?

1. **Circular dependency** — `gpucontext.EventSource` can't reference `gogpu.Window` (gpucontext → gogpu circular import). This means the UI toolkit can never use this callback through the interface.
2. **No rejection** — your callback is fire-and-forget (`func(*Window)`), but every enterprise framework supports rejection (return false / ignore / cancel).
3. **Duplicates existing API** — `Window.SetOnClose(func() bool)` already exists with the right signature, just needs wiring.

## Your tests are valuable

`TestWindowManagerCreateAndClose` and `TestOnWindowCloseCallback` demonstrate real patterns. I'd love to keep the WindowManager test — it covers an untested path.

## Let's design this together

You clearly understand multi-window lifecycle deeply (your #210 report led to ADR-021, and now #213 found another real gap). I'd love to get your perspective on the right architecture here.

**Questions for discussion:**

1. **Per-window `SetOnClose` vs app-level `OnWindowClose`** — for your tabbing project, do you need per-window callbacks (each tab handles its own close), or a centralized "any window closed" observer, or both?

2. **Close rejection** — do you need the ability to CANCEL a close request? (e.g., "unsaved changes, really close?"). The existing `SetOnClose(func() bool)` supports this, but your PR's `func(*Window)` doesn't.

3. **UI toolkit integration** — gogpu/ui subscribes to events through `gpucontext.EventSource` interface. If we add close events there, we can't use `*Window` (circular dependency). The winit/SDL3 pattern uses integer WindowID instead. Does that work for your use case?

4. **Two-phase close** (winit/SDL3 pattern) — `CloseRequested` (before destroy, can reject) + `Closed` (after destroy, cleanup only). Is the two-phase model useful for your tabbing project, or is one phase enough?

Your tabbing use case is exactly the kind of real-world scenario that should drive this API design. What do you think?

lkmavi:
@kolkov  Hi, this isn't the main project, just for illustration purposes :) I liked that you explain in great detail, so I decided to come with an example right away as well. Yes, I'll study this in more detail today and come back with a clear plan. Thank you.

PR #222, opened 2026-05-12
Add a multi-window prototype and close events

lkmavi:
Prototype ready for review — #213 macOS window delegate + multi-window close routing
(For now, linter errors are ignored; I'll definitely fix them later.)

@kolkov Hey! I've pushed the working prototype for issue #213 to the fix/213-add-delegate-single-window-routing branch:

https://github.com/lkmavi/gogpu/tree/fix/213-add-delegate-single-window-routing

What this implements:

Proper close-event detection on macOS — Created a dynamic GoGPUWindowDelegate ObjC class (using objc_allocateClassPair + class_addMethod + ffi.NewCallback, same pattern as GoGPUView) that implements windowShouldClose:. When the user clicks the red close button or presses Cmd+W, the delegate fires, checks the per-window onClose callback, sets shouldClose, and the event loop picks it up.

Multi-window event routing — Each platform window now has its own WindowID, and EventClose carries that ID. The app-level windowCloseEvent handler distinguishes primary vs secondary windows. Closing a secondary window only destroys that window; closing the primary window promotes the next available window (or terminates the app if it's the last one).

Close rejection — SetOnClose(func() bool) on the manager-level Window propagates down to the platform delegate. Returning false from the callback prevents the window from closing.

PlatformWindow interface additions — SetOnClose and Show methods added. Tests and mocks updated for backward compatibility.

ARM64 compatibility — Associated objects (objc_setAssociatedObject) work correctly on Apple Silicon after fixing the ffi argument layout (passing addresses of uintptr variables instead of raw pointers).

Once we finalise and agree on the overall design, I can apply the same pattern to the other platforms (Linux/X11, Wayland, Windows).
For those, I'd implement the platform-specific close event so that SetOnClose and the multi‑window routing behave consistently everywhere.

Would be great to hear if the general direction looks good before extending it further.

### DIFF
--- a/app.go
+++ b/app.go
@@ -555,7 +555,7 @@ func (a *App) windowCloseEvent(event *platform.Event) {
 	if !isPrimary {
 		// Secondary window
 		w := a.windowManager.getByPlatformID(event.WindowID)
-		if w != nil && w.onClose != nil && !w.onClose() {
+		if !w.safeOnClose() {
 			return
 		}
 		if w != nil {
@@ -565,7 +565,7 @@ func (a *App) windowCloseEvent(event *platform.Event) {
 	}
 
 	// Primary window — terminate app
-	if a.primaryWindow != nil && a.primaryWindow.onClose != nil && !a.primaryWindow.onClose() {
+	if !a.primaryWindow.safeOnClose() {
 		return
 	}
 	a.running = false
--- a/app.go
+++ b/app.go
@@ -2,6 +2,7 @@ package gogpu
 
 import (
 	"io"
+	"log"
 	"runtime"
 	"time"
 
@@ -28,10 +29,11 @@ type appRenderLoop interface {
 // This separation ensures the window stays responsive during heavy GPU
 // operations like swapchain recreation.
 type App struct {
-	config     Config
-	manager    platform.PlatformManager // process-level (multi-window)
-	platWindow platform.PlatformWindow  // primary window (per-window ops)
-	renderer   *Renderer
+	config            Config
+	manager           platform.PlatformManager // process-level (multi-window)
+	platWindow        platform.PlatformWindow  // primary window (per-window ops)
+	renderer          *Renderer
+	primaryPlatformID platform.WindowID
 
 	// Multi-thread rendering
 	renderLoop appRenderLoop
@@ -85,6 +87,7 @@ func (a *App) OnDraw(fn func(*Context)) *App {
 	if a.primaryWindow != nil {
 		a.primaryWindow.onDraw = fn
 	}
+
 	return a
 }
 
@@ -184,14 +187,16 @@ func (a *App) Run() error {
 	defer a.manager.Destroy()
 
 	// Create primary platform window.
-	platWindow, err := a.manager.CreateWindow(platform.Config{
-		Title:      a.config.Title,
-		Width:      a.config.Width,
-		Height:     a.config.Height,
-		Resizable:  a.config.Resizable,
-		Fullscreen: a.config.Fullscreen,
-		Frameless:  a.config.Frameless,
-	})
+	platWindow, err := a.manager.CreateWindow(
+		platform.Config{
+			Title:      a.config.Title,
+			Width:      a.config.Width,
+			Height:     a.config.Height,
+			Resizable:  a.config.Resizable,
+			Fullscreen: a.config.Fullscreen,
+			Frameless:  a.config.Frameless,
+		},
+	)
 	if err != nil {
 		return err
 	}
@@ -229,9 +234,13 @@ func (a *App) Run() error {
 
 	// Initialize renderer on render thread (all GPU operations must be on same thread)
 	var initErr error
-	a.renderLoop.RunOnRenderThreadVoid(func() {
-		a.renderer, initErr = newRenderer(a.platWindow, a.config.Backend, a.config.GraphicsAPI, a.config.VSync, a.config.PowerPreference)
-	})
+	a.renderLoop.RunOnRenderThreadVoid(
+		func() {
+			a.renderer, initErr = newRenderer(
+				a.platWindow, a.config.Backend, a.config.GraphicsAPI, a.config.VSync, a.config.PowerPreference,
+			)
+		},
+	)
 	if initErr != nil {
 		return initErr
 	}
@@ -242,19 +251,23 @@ func (a *App) Run() error {
 		// 3. tracker.CloseAll() — auto-tracked resources (LIFO)
 		// 4. onClose callback — manual cleanup (legacy pattern)
 		// 5. Renderer.Destroy() — release GPU device
-		a.renderLoop.RunOnRenderThreadVoid(func() {
-			a.renderer.WaitForGPU()
-			a.renderer.DrainDeferredDestroys()
-			if a.tracker != nil {
-				_ = a.tracker.CloseAll()
-			}
-			if a.onClose != nil {
-				a.onClose()
-			}
-		})
-		a.renderLoop.RunOnRenderThreadVoid(func() {
-			a.renderer.Destroy()
-		})
+		a.renderLoop.RunOnRenderThreadVoid(
+			func() {
+				a.renderer.WaitForGPU()
+				a.renderer.DrainDeferredDestroys()
+				if a.tracker != nil {
+					_ = a.tracker.CloseAll()
+				}
+				if a.onClose != nil {
+					a.onClose()
+				}
+			},
+		)
+		a.renderLoop.RunOnRenderThreadVoid(
+			func() {
+				a.renderer.Destroy()
+			},
+		)
 	}()
 
 	// Register the primary window in the WindowManager.
@@ -276,6 +289,7 @@ func (a *App) Run() error {
 		onResize:   a.onResize,
 		visible:    true,
 	}
+	a.primaryPlatformID = platWindow.ID()
 	a.windowManager.add(a.primaryWindow)
 
 	// Main loop with three rendering modes (ADR-023):
@@ -290,6 +304,7 @@ func (a *App) Run() error {
 	a.invalidator.Invalidate() // Request initial frame
 
 	for a.running && !a.platWindow.ShouldClose() {
+		//log.Printf("[LOOP] running=%v, primaryShouldClose=%v", a.running, a.platWindow.ShouldClose())
 		// Three-mode state detection (ADR-023)
 		continuousRender := a.config.ContinuousRender
 		animating := a.animations.IsAnimating()
@@ -347,6 +362,8 @@ func (a *App) Run() error {
 			a.renderFrameMultiThread()
 		}
 	}
+
+	log.Printf("[EXIT] loop finished. running=%v, primaryShouldClose=%v", a.running, a.platWindow.ShouldClose())
 
 	return nil
 }
@@ -415,9 +432,11 @@ func (a *App) handleSecondaryResize(ev platform.Event) {
 	physW, physH := ev.PhysicalWidth, ev.PhysicalHeight
 	if physW > 0 && physH > 0 {
 		ws := w.surface
-		a.renderLoop.RunOnRenderThreadVoid(func() {
-			ws.resize(physW, physH, a.renderer.device, a.renderer.adapter)
-		})
+		a.renderLoop.RunOnRenderThreadVoid(
+			func() {
+				ws.resize(physW, physH, a.renderer.device, a.renderer.adapter)
+			},
+		)
 	}
 	if w.onResize != nil {
 		w.onResize(ev.Width, ev.Height)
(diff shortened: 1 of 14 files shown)


lkmavi:
This pr will be closed after as it has added comments, hints and sometimes debugging

kolkov (commented):
## Code Review — PR #222

@lkmavi Great prototype! The `windowShouldClose:` delegate approach is exactly right — matches what we validated against SDL3, GLFW, winit, and Qt6. The associated objects pattern for storing `*Window` in ObjC delegates is the correct Apple-standard way. And the move from single `primary` toward per-window routing in `darwinPlatformWindow` is the right architectural direction.

Reviewing this as a prototype per your comment — here's what needs to happen before this can become a merge-ready PR.

---

### Must fix (blocking)

**1. Build broken on Windows/Linux — `PlatformWindow` interface expanded without implementations**

`SetOnClose(func() bool)` and `Show()` were added to `PlatformWindow` interface in `platform.go`, but only `darwinPlatformWindow` implements them. `win32Window` has `var _ PlatformWindow = (*win32Window)(nil)` (line 814) — this static check will fail at compile time. X11 and Wayland will similarly fail when `CreateWindow` tries to return their types as `PlatformWindow`.

**Fix:** Don't add these to the `PlatformWindow` interface. Your delegate calls `win.onClose` directly via the associated object → `*darwin.Window`, so `SetOnClose` on the platform interface is unnecessary for the delegate path. If you need `SetOnClose` propagation from `Window` (window_manager.go) to the platform layer, use an optional interface with type assertion:

```go
type PlatformWindowCloser interface {
    SetOnClose(func() bool)
}

// In Window.SetOnClose:
if closer, ok := w.platWindow.(PlatformWindowCloser); ok {
    closer.SetOnClose(fn)
}
```

Same for `Show()` — `makeKeyAndOrderFront:` is already called in `NewWindow()`, so `Show()` may not be needed at all.

**2. Debug logging in production code**

Three items from the diff:
- `import "log"` added to `app.go`
- `log.Printf("[EXIT] loop finished...")` (active debug line)
- `//log.Printf("[LOOP] running=...")` (commented-out debug line)

Remove all three.

**3. `IsFrameless()` returns hardcoded `false` — regression**

Current code on main:
```go
func (dw *darwinPlatformWindow) IsFrameless() bool {
    w := dw.platform.primary
    w.callbackMu.RLock()
    defer w.callbackMu.RUnlock()
    return w.frameless
}
```

PR replaces with `return false`. This breaks frameless window detection on macOS — any code checking `app.IsFrameless()` will get wrong results.

**Fix:** Since you moved other state from `darwinWindow` to `darwinPlatformWindow`, store `frameless` there too (or query `dw.window` style mask).

**4. `SetHitTestCallback` replaced with no-op — regression**

Current code stores the callback on `darwinWindow.hitTestCallback`. PR replaces with `// no-op`. This breaks custom titlebar hit-testing on macOS, used by frameless windows for drag/resize regions.

**Fix:** Store `hitTestCallback` on `darwinPlatformWindow` and wire it as before.

**5. Secondary window close events won't reach the event loop**

`PollEvents()` calls `p.primary.pollEvents(p.app)` — only the primary window's event queue is polled. When the delegate fires `windowShouldClose:` on a secondary window, it sets `shouldClose=true` on that secondary `darwin.Window`. But `pollEvents` only checks `p.primary.window.ShouldClose()` and queues `EventClose` with the primary's `WindowID`.

Result: secondary window close events from the delegate **are never delivered** to the app event loop. The delegate fires, `shouldClose` is set on the secondary, but nobody reads it.

This is the core multi-window routing problem you identified in your #213 comment. For the production PR, `PollEvents` needs to iterate all windows (or use a shared event channel).

**6. Primary window promotion — renderer not updated**

When primary is closed and a secondary is promoted:
```go
a.primaryWindow = newPrimary
a.primaryPlatformID = newPrimary.platformID
a.platWindow = newPrimary.platWindow
```

But `a.renderer` was initialized with the OLD primary's `PlatformWindow`. After promotion, `a.renderer.platWindow` still references the destroyed window. `renderer.beginFrame()` calls `r.platWindow.GetHandle()` — crash or undefined behavior on next frame.

Also `a.renderer.primary` (the `windowSurface`) holds the old wgpu surface created from the old window's native handle.

**Recommendation:** Defer primary promotion to a future PR. It requires renderer surface migration which is a separate problem. For now, closing the primary should terminate the app (current behavior), and only secondary windows get per-window close routing via the delegate.

**7. Duplicate comment on `windowCloseEvent`**

```go
// windowCloseEvent handles CloseEvent from the platform.
// windowCloseEvent handles CloseEvent from the platform.
func (a *App) windowCloseEvent(event *platform.Event) {
```

Remove the duplicate line.

**8. Race condition: duplicate `if p.primary == nil` check in `CreateWindow`**

```go
func (p *darwinPlatform) CreateWindow(config Config) (PlatformWindow, error) {
    ...
    if p.primary == nil {     // ← NO LOCK — race condition
        p.primary = w
    }
    ...
    p.mu.Lock()
    if p.primary == nil {     // ← under lock — correct
        p.primary = w
    }
    p.mu.Unlock()
```

The first check without lock is a data race. Keep only the locked version.

---

### Suggestions (non-blocking)

**9. Test `TestApp_CloseSecondaryWindowRecyclesID` changed incorrectly**

PR replaces `app.closeSecondaryWindow(secID)` with manual `remove()` + `release()` + `sec.visible = false`. This bypasses `closeSecondaryWindow()` entirely — skipping GPU surface cleanup, `platWindow.Destroy()`, and `onAnyWindowClosed` callback. The original call is nil-safe (`if w.platWindow != nil`), and the PR already adds `platWindow: &mockWindow{}` to the test. Keep using `closeSecondaryWindow()`.

**10. ~380 lines of re-indent noise in `selectors.go`**

The entire `selectors.go` diff (+197/-185) is reformatting — moving closing parens to new lines. The only real change is adding 1 line: `selectors.windowShouldClose`. Same pattern in `app.go` (RunOnRenderThreadVoid blocks, append calls). This makes review significantly harder. For the production PR, please don't reformat code that isn't functionally changed.

**11. `findNewPrimaryWindowID` breaks `WindowManager` encapsulation**

```go
func (a *App) findNewPrimaryWindowID() InternalWindowID {
    a.windowManager.mu.RLock()         // accesses private lock
    ...
    for _, id := range a.windowManager.order {  // accesses private slice
```

Consider a method like `windowManager.firstOtherID(exclude InternalWindowID) InternalWindowID`.

**12. Best practice: `setDelegate:nil` before destroying delegate**

In `Destroy()`, you clear the associated object and release the delegate, but the delegate is still attached to the NSWindow. If Cocoa sends any callback between those steps, the delegate's `windowShouldClose:` would fire. Your code handles this safely (returns 1 if association is nil), but the standard pattern is to call `nsWindow.setDelegate(nil)` first, then release the delegate. This prevents any future callbacks.

---

### Architecture — summary

The **delegate + associated objects** approach is production-ready. `GoGPUWindowDelegate` follows exactly the `GoGPUView` pattern already in our codebase. The `windowShouldClose:` → `onClose` callback → `shouldClose` flow is correct.

For the production PR, I'd recommend scoping it to:

1. **Delegate + associated objects** — keep as-is ✅
2. **`SetOnClose` on `darwin.Window`** — keep ✅
3. **Per-window `darwinPlatformWindow` with own `*darwin.Window`** — keep, but restore `IsFrameless`/`SetHitTestCallback`
4. **Primary promotion** — defer to a future PR (requires renderer surface migration)
5. **Multi-window `PollEvents`** — defer to a future PR (requires iterating all windows)
6. **`PlatformWindow` interface** — don't expand; use type assertions

This way the PR delivers the delegate infrastructure without regressions, and multi-window routing builds on top of it.

Looking forward to the next iteration! 🎯

lkmavi:
@kolkov I want to check everything again and I will write when everything is finished. For now I have done the following:
1. fixed
2. delete
3. fixed
4. fixed
5. fixed
6. Create https://github.com/gogpu/gogpu/issues/223
7. fixed
8. fixed
9. fixed
10.Sorry:( I saved this for the future, the rule worked when pushing :( It shouldn't happen again in the future
11. Removed
12. fixed

lkmavi:
@kolkov Added tests. Done. 

P.S I have moved point 6 into a separate issue, and I am ready to get started on it.

kolkov (commented):
## Code Review v2 — PR #222

@lkmavi Great progress! Most v1 issues are addressed. CI green (Build/Test/Lint on all 3 platforms). Here's the validated review.

### Status of v1 issues

| # | Issue | Status | Notes |
|---|-------|--------|-------|
| 1 | PlatformWindow interface | ✅ Fixed | `PlatformWindowCloser` optional interface |
| 2 | Debug logging | ✅ Fixed | No `import "log"` or `log.Printf` |
| 3 | IsFrameless regression | ⚠️ **Partially fixed** | See new issue #1 below |
| 4 | SetHitTestCallback regression | ✅ Fixed | `dw.hitTestCallback` + `dw.callbackMu` |
| 5 | Secondary close events | ✅ Fixed | `PollEvents` iterates `p.windows` |
| 6 | Primary promotion | ✅ Deferred | Issue #223 created |
| 7 | Duplicate comment | ✅ Fixed |
| 8 | Race in CreateWindow | ✅ Fixed | Single locked block |

### Must fix

**1. `SetFrameless` / `IsFrameless` write/read mismatch**

`SetFrameless` now writes to `dw.frameless`:
```go
func (dw *darwinPlatformWindow) SetFrameless(frameless bool) {
    dw.frameless = frameless  // writes to darwinPlatformWindow
    ...
}
```

But `IsFrameless` was NOT changed — it still reads from `dw.platform.primary`:
```go
func (dw *darwinPlatformWindow) IsFrameless() bool {
    w := dw.platform.primary         // reads from primary's darwinWindow!
    w.callbackMu.RLock()
    defer w.callbackMu.RUnlock()
    return w.frameless
}
```

Result: `SetFrameless(true)` on any window → `IsFrameless()` returns false (still reading primary's value).

Fix:
```go
func (dw *darwinPlatformWindow) IsFrameless() bool {
    return dw.frameless
}
```

**2. Nil dereference order in `windowCloseEvent`**

```go
w := a.windowManager.getByPlatformID(event.WindowID)
if !w.safeOnClose() {    // w can be nil
    return
}
if w != nil {            // nil check AFTER method call
```

`safeOnClose` handles nil receiver, so this works but is a Go antipattern. Standard approach:

```go
w := a.windowManager.getByPlatformID(event.WindowID)
if w == nil {
    return
}
if !w.safeOnClose() {
    return
}
a.closeSecondaryWindow(w.id)
```

**3. `CreateWindowDelegate(nil)` — missing nil guard**

`CreateWindowDelegate` doesn't check `win == nil`. `TestWindowDelegateCreationNilWindow` expects an error but the function would return `(delegate, nil)`. Test passes CI only because all darwin tests are skipped via `TestMain` → `os.Exit(0)` when `CI=true`.

Fix: add `if win == nil { return 0, ErrWindowCreationFailed }` at the top of `CreateWindowDelegate`.

### Should fix

**4. `LogicalSize` fallback reads primary's config**

```go
func (dw *darwinPlatformWindow) LogicalSize() (int, int) {
    if dw.window != nil {
        return dw.window.Size()
    }
    return dw.platform.primary.config.Width, dw.platform.primary.config.Height
}
```

For secondary windows this returns primary's size. `PhysicalSize` returns `(0, 0)` in the same case — inconsistent. Since `dw.window` is always non-nil after `CreateWindow`, the fallback should never fire, but if it does — `return 0, 0` is safer.

**5. `Show()` on `darwinPlatformWindow` — dead code**

`Show()` is implemented but never called. In v1, `NewWindow` in `window_manager.go` called `w.platWindow.Show()` — that was removed. The method exists on `darwinPlatformWindow` and mocks, but nothing invokes it. Either wire it in or remove it to keep the API clean.

### Suggestions (non-blocking)

**6. `PollEvents` uses `Lock` where `RLock` suffices** — `p.mu` protects `p.windows` slice, PollEvents only reads it. Individual window events are protected by `w.eventMu`.

**7. `safeOnClose` panic = close rejected** — when `onClose` panics, close is silently rejected. Consider whether panic should mean "allow close" (the callback is broken). Either way, add a one-line comment documenting the choice.

### Architecture: Universal App Lifecycle (ADR-026)

Your PR and issue #223 prompted deep research into 15 enterprise frameworks (Qt6, GTK4, SDL3, winit, GLFW, Electron, AppKit, Win32, Flutter, React Native, MAUI, Bevy + wgpu source code).

**Key finding:** gogpu is a GPU rendering ecosystem — we render into surfaces provided by the platform, similar to how Flutter renders into `FlutterView`. The "primary window" concept is desktop-only and doesn't generalize to mobile (Android Activity lifecycle), web (Canvas + tab visibility), or headless (compute-only). No cross-platform framework that supports mobile has a "primary window" concept.

**All 15 frameworks converge on the same universal pattern:**
- GPU Device is global, surfaces are per-window/view (nilable — dropped on Android suspend, recreated on resume)
- App lifecycle (running/suspended) is orthogonal to surface lifecycle (create/destroy)
- Desktop is a trivial subset of the mobile lifecycle (surfaces always available, suspend never fires)
- ONE universal mode for all platforms, not separate desktop/mobile code paths

**What this means for this PR:** Your direction is architecturally correct. Per-window routing and decoupling from `primary` is exactly where we need to go. We've created **ADR-026: Universal App Lifecycle** and a phased implementation plan:

| Phase | What | Who |
|-------|------|-----|
| **1** | macOS delegate (this PR) | @lkmavi |
| **2** | Renderer decoupling — Device/Queue independent of any window | @kolkov |
| **3** | Universal lifecycle API — `AppLifecycle` states, `QuitOnLastSurfaceClosed` | @kolkov |
| **4** | Mobile platforms — Android `ANativeWindow`, iOS `CAMetalLayer`, Web canvas | community |

Phase 1 (your PR) is the foundation. After merge, we'll decouple the renderer from the "primary window" concept entirely (Phase 2), then build the universal lifecycle API on top (Phase 3). This prepares gogpu for Android, iOS, and WASM support with zero architectural rewrites later.

If you're interested in iOS platform work in the future (Phase 4.2) — your macOS/ObjC expertise would be a perfect fit.

### Summary

Solid improvement from v1. The delegate infrastructure is production-ready. Fix items 1-3 and this is merge-ready.

kolkov (approved):
## Approved — all review items resolved

@lkmavi All 3 must-fix items from review v2 are fixed:

1. ✅ `IsFrameless()` — now reads `dw.frameless` (was reading `primary.frameless`)
2. ✅ Nil check order — `w == nil` before `safeOnClose()`
3. ✅ `CreateWindowDelegate(nil)` — nil guard added
4. ✅ Bonus: `LogicalSize` fallback returns `(0, 0)` (consistency with `PhysicalSize`)
5. ✅ CI fix: darwin-only test removed from cross-platform file

CI green: Build ✅ 3/3, Test ✅ 3/3, Lint ✅, Formatting ✅.

**Note for the record:** Linux X11 and Wayland currently emit `EventClose` without `WindowID` (always 0). This is a pre-existing issue on main — not introduced by this PR. The old fallback `event.WindowID == 0 → isPrimary` masked it but didn't fix close rejection (the main loop's `ShouldClose()` check exits regardless). We'll fix this in Phase 2 when unifying all platforms to include proper `WindowID` in every event. Tracked in FEAT-LIFECYCLE-001.

Great work on this PR — the delegate infrastructure is solid and production-ready. 🎯

Issue #240, opened 2026-05-16
Holding down the backspace key in the textfields does not repeat


celer:
Tested on linux

Expected Behavior:

Write some text into a text field, then hold down the backspace key, the characters are continously deleted at a pace of about 500ms a character

Actual Behavior

The backspace key must be pressed repeatedly for each character that is deleted. 

This behavior is apparent in the TextField in the input section of the gallery example app.

celer:
The situation is improved, the key repeat works, but now there is an update problem, when I hold down a key sometimes (it doesn't always happen) the GUI stops updating so it will jump from one character 'd' to 20 or so characters 'd' without any gui updates in between. So I as a user get the experiance that the text suddenly jumps

This is tested on linux.

PR #260, opened 2026-05-19
fix: Wayland key repeat — timerfd replaces goroutine (#240, v0.38.0)

kolkov:
Goroutine never called WakeUp → GUI freeze. timerfd in Poll set (GLFW/winit pattern). Zero goroutines, zero data races.


kolkov:
Superseded by PR #261 (merged as v0.37.10). This was the v0.38.0 branch before version correction to PATCH.

PR #265, opened 2026-05-21
fix/264-add-custom-menu : Add custom menu

lkmavi:
@kolkov So, I added the ability to add an additional menu. I also had the idea to add functionality for deleting the standard menu. For example, if we want to remove and add only our own items, then the CLEAR method would be useful.

Снимок экрана 2026-05-21 в 20 09 32

Unrelated to the task, but I'm getting an error with smoke tests.

Perhaps this is just me?

starting TestDarwinAppRunPanicOnDrawNoTriangle
2026/05/21 19:59:48 INFO adapter selected name="Apple M3 Max" type=DiscreteGPU
panic: forced draw panic

### DIFF
--- a/app.go
+++ b/app.go
@@ -71,6 +71,7 @@ type App struct {
 	primaryWindow *Window
 
 	menu                   *Menu
+	customMenus            map[string]*Menu
 	pendingSystemMenuItems map[SystemMenu][]MenuItem
 }
 
@@ -308,8 +309,8 @@ func (a *App) initPlatform() (platform.PlatformWindow, error) {
 	}
 
 	// Apply any menu set before Run().
-	if a.menu != nil {
-		a.SetMenu(a.menu)
+	if a.menu != nil || len(a.customMenus) > 0 {
+		a.updateNativeMenu()
 	}
 	// Apply pending system menu items.
 	if a.pendingSystemMenuItems != nil {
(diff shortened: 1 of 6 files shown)


kolkov:
@lkmavi Good work on custom menus! Architecture is solid — `updateNativeMenu()` merging main + custom items, recursive `convertMenuToPlatformItems` for submenus, `addPlatformItem` with proper separator/role/submenu handling.

CI all green, tests cover the custom menu flow well. A few items before merge:

**1. BREAKING: `NewMenu()` → `NewMenu(title)` — needs migration note**

`NewMenu()` was public API (used in README, examples). Changing signature to `NewMenu(title string)` breaks existing users. Two options:
- (a) Keep `NewMenu()` with empty title, add `NewMenuWithTitle(title)` — backward compatible
- (b) Keep the break, update README example (line 326: `gogpu.NewMenu().` → `gogpu.NewMenu("").`)

I'd prefer (a) for backward compat, but (b) is acceptable since we're pre-1.0. If (b), please update `README.md` line 326.

**2. Map iteration order — custom menus are unordered**

```go
for _, menu := range a.customMenus {
```

Go map iteration is random. If user adds "File", "Edit", "View" custom menus, they appear in random order each run. Consider `[]struct{name string; menu *Menu}` or `orderedMenus []*Menu` to preserve insertion order.

**3. `AddCustomMenu` vs `UpdateCustomMenu` — identical implementation**

Both do exactly the same thing:
```go
func (a *App) AddCustomMenu(name string, menu *Menu) { ... }
func (a *App) UpdateCustomMenu(name string, menu *Menu) { ... }
```

Either merge into one (`SetCustomMenu`) or add semantics (e.g., `AddCustomMenu` errors if name exists, `UpdateCustomMenu` errors if not found).

**4. Example removed SetMenu showcase**

The example now only shows custom menus but removed the `SetMenu` demo (About, Preferences, Quit roles). Consider keeping both — SetMenu for the app menu + AddCustomMenu for additional menus. The old example was a good showcase of role-based items.

**5. Minor: `Submenu *Menu` in `MenuItem` — recursive type, nice**

The recursive `MenuItem.Submenu *Menu` → `convertMenuToPlatformItems` recursion is clean. No depth limit needed since macOS enforces its own.

Everything else looks great — the `addPlatformItem` refactor in `platform_darwin.go` is much cleaner than the original flat loop.

kolkov (approved):
All 4 review items fixed: NewMenu() backward compat, ordered slice, SetCustomMenu upsert, example restored. Clean work!

PR #281, opened 2026-05-29
feat/adr-36 : Fix on after rewiew

lkmavi:
1. fix Lines 282 and 303 — uintptr → unsafe.Pointer via syscall.SyscallN. Nolints are needed, so they are left in.
Снимок экрана 2026-05-29 в 15 41 27

2. fix const
3. fix add const
4. migration setAllowedContentTypes:
5. runtime.KeepAlive instead of _ = backing
6. The HRESULT for IShellItemArray.GetCount was not checked.Status:  Now returns an error on failure.

### DIFF
--- a/internal/platform/dialog_darwin.go
+++ b/internal/platform/dialog_darwin.go
@@ -91,20 +91,28 @@ func applyDialogCommon(panel darwin.ID, opts FileDialogOptions) {
 		}
 	}
 
+	// Filters are skipped in directory mode: the panel restricts to folders, not file types.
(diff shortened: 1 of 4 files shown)


lkmavi:
@kolkov 
If everything is fine, then after the merge I take the 2nd step into work)

kolkov (approved):
All review comments addressed. UTType migration, runtime.KeepAlive, GetCount HRESULT check, naming fixes, isOpen guard — all correct. Great turnaround!

PR #283, opened 2026-05-31
feat/adr-36 : Add Linux zenity/kdialog + D-Bus portal 

lkmavi:
Summary
---

- Added internal/platform/dialog_linux.go (stubs)

Test plan
---
- [x] go build ./... passes
- [x] go test ./...  passes
- [x] golangci-lint run — 0 issues
- [x] running the example on Ubuntu 26.04(ARM)
- [x] running the example on Ubuntu 26.04 Wayland (ARM)

Demo
---
x11
---

https://github.com/user-attachments/assets/1cf5cc4a-3041-447c-8e0a-676ded257529

Wayland
---

https://github.com/user-attachments/assets/770b9e26-a9e5-41a1-8bab-19f1852d9e7b


kolkov (approved):
## Code Review: Linux File Dialogs — D-Bus + zenity/kdialog (ADR-036 Phase 2)

### Overall: APPROVE ✅

@lkmavi Outstanding work. 1365 LOC hand-written D-Bus wire protocol + 990 LOC tests, zero external dependencies. This is exactly the kind of engineering that sets gogpu apart. Smart fallback (portal → zenity → kdialog), correct SASL EXTERNAL auth, proper wire format, and comprehensive tests.

### Strengths

- **D-Bus wire protocol done right** — correct header layout, string/signature encoding, alignment rules, little-endian byte order. All per spec.
- **Smart fallback logic** — `errPortalUnavailable` distinguishes "can't reach portal" (safe fallback) from "portal reached but error mid-dialog" (don't fall back → no double dialog). Subtle UX win.
- **SASL EXTERNAL auth correct** — NUL + `AUTH EXTERNAL <hex-uid>` + `BEGIN`. Hex encoding of UID as ASCII-hex-of-decimal verified.
- **Security** — `exec.Command` argv (no shell injection), Unix socket only (rejects TCP), message size limits, `url.PathUnescape` for URIs ✅
- **990 LOC tests** — fake binary via `writeFakeBinary` (brilliant), D-Bus encoding round-trips, portal response decoding, alignment verification, table-driven zenity/kdialog arg tests

### Suggestions for Follow-up (not blocking merge)

**1. `AddMatch` for portal Response signal** (`dialog_linux.go`, after `sendCall`):
The code waits for `Response` signal but never registers a match rule via `org.freedesktop.DBus.AddMatch`. Works on standard dbus-daemon (implicit delivery), but may fail on dbus-broker (Fedora/RHEL 9+ default). Low risk now, but worth adding.

**2. Timeout on `waitResponse` / `hello`** — both loop forever with no deadline. If portal crashes → app freezes. Suggestion: `conn.SetDeadline(time.Now().Add(5*time.Minute))` for dialog, 5s for hello handshake.

**3. Signal ordering** — if Response signal arrives *before* METHOD_RETURN (D-Bus spec allows this), it's dropped and `waitResponse` blocks forever. Suggestion: buffer signals during phase 1, check on entering phase 2.

**4. `InitialDirectory` ignored in portal path** — zenity uses `--filename=<dir>/`, kdialog uses start dir, but `encodeFileChooserBody` doesn't encode `current_folder` option. Asymmetry with subprocess backends.

**5. `dbusTypeAlign` returns 8 for variant `'v'`** (line 1216) — per D-Bus spec, variant alignment = 1, not 8. Structs `(` and dict entries `{` = 8, but variant = 1. Currently latent (no `av` arrays parsed), but will bite when D-Bus is reused for ADR-040 menus.
```go
// Current (bug):
case 'v', '(', '{':
    return 8

// Fix:
case '(', '{':
    return 8
case 'v':
    return 1
```

### Minor

- Line 345: `ADR-XX` → should be `ADR-036`
- `reflect.DeepEqual` for `[]string` → `slices.Equal` (Go 1.21+, optional)
- `dbusReadLine` has no buffer limit — add `if len(buf) > 4096 { return err }` guard
- `dbusParseParams` doesn't percent-decode values (spec requires it, rarely matters)
- kdialog space-separated output: note `--separator` flag exists in KDE Frameworks 5.x+

### Questions

1. **dbus-broker tested?** Fedora 9+ uses it as default. Does portal signal arrive without `AddMatch`?
2. **Portal `current_folder` omission** — intentional or oversight?
3. **File splitting** — 1365 LOC is substantial. Extract D-Bus wire protocol to `dbus_linux.go` now (for ADR-040 reuse) or later?

### Security: LOW RISK ✅

No command injection, local Unix socket only, message size limits, proper URI escaping.

Merge as-is, track suggestions as follow-up issues. Exceptional contribution! 🎉

kolkov:
@lkmavi Before we merge, could you fix two items in this PR?

**1. Variant alignment bug** (`dialog_linux.go:1216`):
```go
// Current (bug):
case 'v', '(', '{':
    return 8

// Fix — variant alignment is 1 per D-Bus spec:
case '(', '{':
    return 8
case 'v':
    return 1
```

**2. `InitialDirectory` in portal path** (`encodeFileChooserBody`):
zenity and kdialog both use `InitialDirectory`, but portal path doesn't encode `current_folder`. Could you add it to the `a{sv}` dict? Format: `current_folder` → variant `ay` (path bytes + NUL terminator).

Everything else from the review is follow-up — these two are quick fixes to keep the merge clean. Thanks!

lkmavi:
@kolkov Hi, thanks :) I'll be back with some changes to this MR.

lkmavi:
@kolkov

**Fixed:**

- **Suggestion 2 — Timeouts:** Both `hello` and `waitResponse` are now deadline-guarded. `hello` uses a 5-second deadline (`c.rw.SetDeadline(time.Now().Add(5 * time.Second))`), cleared afterwards; `waitResponse` uses a 5-minute deadline to cover the user interaction window. A hung or crashed portal will now unblock with a read error.
- **Suggestion 5 — `dbusTypeAlign` variant bug:** Fixed `'v'` to return `1` instead of `8` per spec, with an explanatory comment. `'('` and `'{'` remain `8`.
- **Suggestion 4 — `current_folder` omission:** Confirmed it was an oversight — `current_folder` is now encoded in `encodeFileChooserBody` as `ay` (NUL-terminated path per xdg-portal spec), making the portal path consistent with zenity/kdialog.
- **Minor — `dbusReadLine` buffer limit:** Added a 4096-byte guard.
- **Minor — `slices.Equal`:** Replaced `reflect.DeepEqual` for `[]string` comparisons in tests.
- **Question 3 — D-Bus file split:** Done — the wire protocol lives in `dbus_linux.go`, reusable for ADR-040.

**Still open (tracked for follow-up):**

- **Suggestion 1 — `AddMatch` for dbus-broker:** Tested on Ubuntu (works without `AddMatch`). Fedora testing hit virtualization/Wayland issues on ARM — need to retest on live hardware before confirming dbus-broker behavior.

https://github.com/user-attachments/assets/091505b0-72ad-45df-9572-ae9560792597


- **Suggestion 3 — Signal ordering:** Fixed waitResponse in dbus_linux.go now buffers the Response signal if it arrives before METHOD_RETURN:

kolkov (approved):
All review fixes confirmed — variant alignment, current_folder, timeouts, signal ordering, file split, buffer limit, slices.Equal. CI 10/10. Excellent turnaround @lkmavi!

kolkov:
@lkmavi Please rebase on main — we merged the clear color fix (#284) after your PR was created. `git fetch origin main && git rebase origin/main && git push -f`

PR #290, opened 2026-06-03
Adr 40 linux dbus amenu

lkmavi:
1. serve() goroutine relies on conn.rw.Close() to break readMsg(). Correct for net.Conn blocking I/O. stopCh declared but only used as close flag — no functional issue.
--Fixed
2. menuEncodeCall for RegisterWindow doesn't set NO_REPLY_EXPECTED flag. The reply arrives in serve() and gets silently skipped at msg.Path != objPath check. Harmless — no hang, no leak.
--Fixed
3. menuAssembleMsg has similar structure to dbusEncodeMsg — potential consolidation point when third D-Bus consumer appears (notifications?). Not worth refactoring now with only two consumers. 
--Fixed

### DIFF
--- a/internal/platform/dbus_linux.go
+++ b/internal/platform/dbus_linux.go
@@ -479,6 +479,30 @@ func (b *msgBuf) arrayEnd(lenFieldPos, contentPos int) {
 	binary.LittleEndian.PutUint32(b.data[lenFieldPos:], uint32(len(b.data)-contentPos))
 }
 
+// dbusAssembleMsg assembles a complete little-endian D-Bus message from a
+// pre-built header-fields slice and a body. flags is written to byte 2 of the
+// fixed header (e.g. dbusFlagNoReplyExpected = 0x02; pass 0 when unused).
+// Used by both the generic D-Bus client (dbus_linux.go) and the dbusmenu
+// server (menu_linux.go) so the wire-encoding logic lives in exactly one place.
+func dbusAssembleMsg(msgType, flags byte, serial uint32, hdrBytes, body []byte) []byte {
+	var fixed [16]byte
+	fixed[0] = 'l' // little-endian
+	fixed[1] = msgType
+	fixed[2] = flags
+	fixed[3] = 1 // protocol version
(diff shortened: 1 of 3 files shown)


kolkov (changes requested):
## Code Review: D-Bus Follow-up Fixes

### Verdict: REQUEST CHANGES — 1 bug

**BUG: Wrong flag value** (`menu_linux.go:38`):
\`\`\`go
// Current (bug):
dbusFlagNoReplyExpected byte = 0x02 // this is NO_AUTO_START!

// Fix:
dbusFlagNoReplyExpected byte = 0x01 // NO_REPLY_EXPECTED per D-Bus spec
\`\`\`

Per [D-Bus spec](https://dbus.freedesktop.org/doc/dbus-specification.html) and [godbus/dbus](https://github.com/godbus/dbus/blob/master/message.go):
- \`0x01\` = NO_REPLY_EXPECTED
- \`0x02\` = NO_AUTO_START

Setting \`0x02\` tells the bus NOT to auto-start the registrar service. On socket-activated DEs this could silently prevent menus from appearing.

### Everything else: PASS ✅

- **\`dbusAssembleMsg\` DRY refactor** — clean extraction, wire-compatible, file dialogs unaffected
- **\`stopCh\` restoration** — legitimate improvement over our v0.41.1 removal. Distinguishes intentional shutdown from unexpected connection loss. Race-free (\`close(stopCh)\` before \`conn.rw.Close()\`).
- **v0.41.1 fixes preserved** — registerSerial tracking, logger().Debug writes, ErrorName parsing all intact
- **Tests** — 6 new tests, good coverage. Update \`dbusFlagNoReplyExpected\` constant and tests auto-pass.

One-line fix, then ready to merge! Great follow-up work @lkmavi.

lkmavi:
Thanks for the shoutout! And sorry about the oversight with the NO_REPLY_EXPECTED flag — got a bit carried away fixing a merge conflict and it slipped through. All sorted now though

kolkov (approved):
Flag value fixed: 0x02 → 0x01 (NO_REPLY_EXPECTED). Verified against D-Bus spec. All 10/10 CI pass. LGTM!

PR #299, opened 2026-06-08
fix/292 : Fix multiwindow

lkmavi:
1. macOS — only 1 window shown (window_manager.go:354): NewWindow() was creating the platform window but never calling    
  platWindow.Show(). The v0.41.5 "hidden-then-show" refactor moved Show() out of CreateWindow(), but only wired it up in the
  primary window path (app.go). Fixed: platWindow.Show() is now called after GPU surface configuration in NewWindow().      
  2. Wayland — first window infinite loading (libwayland_xdg.go, libwayland_input.go): Global callback handles              
  (xdgCallbackHandle, inputCallbackHandle) were overwritten when the second window's LibwaylandHandle was created, so the   
  first window's xdg configure acks were sent to the wrong surface — the compositor never mapped the first surface. Fixed:  
  replaced both globals with per-proxy maps (xdgSurfaceHndls, xdgWmBaseHndls, toplevelHandles) keyed by C proxy pointer.    
  3. Wayland — closing one window closes both (platform_linux.go): p.primaryWindowID was overwritten to the secondary       
  window's ID on the second CreateWindow() call, and ShouldClose() shared state across windows. Fixed: secondary windows get
  independent Wayland connections via createSecondaryConn(), with their own shouldClose flag; ShouldClose() and Destroy()   
  dispatch to the secondary's own state when w.secondary != nil.                                                            
  4. Wayland — secondary events missed during blocking wait (platform_linux.go:WaitEvents): WaitEvents() only polled the    
  primary connection's FD. Fixed: secondary connection FDs are appended to the unix.Poll set so the main loop unblocks for  
  any window's compositor events. 
  
  
 @kolkov remove replay before merge

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,8 @@ require (
 	golang.org/x/sys v0.45.0
 )
 
+replace github.com/gogpu/wgpu => github.com/lkmavi/wgpu v0.0.0-20260608204626-7362577e873b
+
 require (
 	github.com/go-webgpu/webgpu v0.5.2 // indirect
 	github.com/gogpu/naga v0.17.14 // indirect
--- a/go.mod
+++ b/go.mod
@@ -10,8 +10,6 @@ require (
 	golang.org/x/sys v0.45.0
 )
 
-replace github.com/gogpu/wgpu => github.com/lkmavi/wgpu v0.0.0-20260608204626-7362577e873b
-
 require (
 	github.com/go-webgpu/webgpu v0.5.2 // indirect
 	github.com/gogpu/naga v0.17.14 // indirect
--- a/go.sum
+++ b/go.sum
@@ -6,9 +6,9 @@ github.com/gogpu/gpucontext v0.19.0 h1:cVm5wUtK7F/anyYXc7vncrB/F9ujBQ2XUbi3oq4hB
 github.com/gogpu/gpucontext v0.19.0/go.mod h1:6zwdmYXH5GQltoiHbb3WXVS/UJ5bFsCux0mXCVqGlzY=
 github.com/gogpu/gputypes v0.5.0 h1:i2ED/9w6m6yLxf8XJT69/NIMSNTLO2y5F1LqvugCKIE=
 github.com/gogpu/gputypes v0.5.0/go.mod h1:cnXrDMwTpWTvJLW1Vreop3PcT6a2YP/i3s91rPaOavw=
-github.com/gogpu/naga v0.17.14 h1:/xUBCkCHSaSi991SSOeYADS9jjlAQivea4brYduI/So=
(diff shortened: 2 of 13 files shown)


kolkov (commented):
@lkmavi Enterprise code review of PR #299 (multiwindow fix). Great work — 4 real bugs fixed.

### Architecture: ✅ Sound

Independent Wayland connections per secondary window (`secondaryWaylandConn`) is the right approach. Each window gets its own `wl_display` + `wl_surface` + `xdg_toplevel`. Input stays on primary seat (comment at line 184). This matches how GLFW handles multi-window Wayland.

### Per-proxy callback routing: ✅ Correct

Replacing global `xdgCallbackHandle` / `inputCallbackHandle` with per-proxy maps (`xdgSurfaceHndls`, `xdgWmBaseHndls`, `toplevelHandles`) keyed by C proxy pointer — clean solution. Mutex-protected, unregistered in `Close()`.

### Thread panic propagation: ✅ Nice catch

`threadResult` with `recover()` in `Call`/`CallVoid` re-panics on caller. Prevents silent goroutine death on render thread.

### macOS GetPoint buffer: ✅ Correct

`[2]float64` → `[4]float64` for goffi HFA return — matches Go 1.26 checkptr with `-race`.

### Show() in NewWindow: ✅ Our regression fixed

`platWindow.Show()` added in `window_manager.go:354` — fixes the hidden-then-show regression from v0.41.5 where we added Show() for primary window but forgot secondary.

### Items to address before merge:

1. **`replace` directive in go.mod** (line 8) — `replace github.com/gogpu/wgpu => github.com/lkmavi/wgpu ...` must be removed. Please rebase on `main` which now has wgpu v0.29.13 + naga v0.17.14 (v0.41.8 just released).

2. **PR body says "remove replay before merge"** — is there debug/replay code still in the PR that needs cleanup?

3. **No CSD for secondary windows** — `createSecondaryConn` passes `decorName/decorVersion` to `OpenLibwayland` for SSD. Without SSD, secondary windows have no decorations. This is a reasonable first implementation, but worth a brief comment in `createSecondaryConn` noting that CSD for secondaries is deferred (separate subsurface + pointer routing per window is complex).

4. **Defensive: `WaitEvents` secondary FD iteration** (line 2566) — `secs := p.secondaries` copies slice header under RLock, then iterates after RUnlock. Currently safe (main thread is single-threaded), but a brief comment noting the safety assumption would help future readers.

5. **naga v0.17.13 in go.sum** — resolves after rebase on v0.41.8.

### Summary

Solid multiwindow implementation. Main blocker: remove `replace` directive and rebase on v0.41.8 (wgpu v0.29.13 + naga v0.17.14). Rest are minor suggestions. CI 9/9 pass.

lkmavi:
@kolkov fixed on after rewiew

PR #309, opened 2026-06-14
fix/300 : Fix artefact tfrom GNOME : Fix hiden bar macos

lkmavi:
- darwin/window.go — setLayer before setWantsLayer:YES → fixes invisible title bar on macOS (layer-hosting vs. layer-backed mode)
- libwayland_input.go — forced csdPendingResize on state change → maximize→fullscreen handling at the same resolution
- libwayland_csd.go — shouldDestroy → shouldHide, surfaces remain alive → fixes GNOME artifact: null state is applied atomically by the parent commit
- libwayland_xdg.go — csdPendingRepaint = false moved after ResizeCSD → allows state change without resizing to bypass early return

### DIFF
--- a/internal/platform/darwin/window.go
+++ b/internal/platform/darwin/window.go
@@ -332,11 +332,13 @@ func (w *Window) SetMetalLayer(layer ID) {
 		return
 	}
 
-	// Enable layer backing
-	w.contentView.SendBool(selectors.setWantsLayer, true)
-
-	// Set the layer
+	// Layer-hosting mode: setLayer: FIRST, then setWantsLayer:YES.
+	// Apple docs require this exact order to enter layer-hosting mode.
+	// Reversing the order creates layer-backed mode where AppKit manages
+	// the layer lifecycle — on macOS 15+ this causes the Metal content
+	// to composite above the title bar, making it invisible.
 	w.contentView.SendPtr(selectors.setLayer, layer.Ptr())
+	w.contentView.SendBool(selectors.setWantsLayer, true)
 
 	w.metalLayer = layer
 }
--- a/internal/platform/platform_linux.go
+++ b/internal/platform/platform_linux.go
@@ -852,8 +852,15 @@ func (p *waylandPlatform) initSingleConnection(config Config) error { //nolint:g
 		}
 	}
 
-	// Activate CSD if SSD was not available and window is not frameless
-	if decorGlobal == nil && !config.Frameless {
+	// Activate CSD if SSD was not negotiated and window is not frameless.
+	// Two cases require CSD:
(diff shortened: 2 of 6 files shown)


kolkov (approved):
Excellent work @lkmavi! Thorough code review completed — every change verified against protocol specs, Apple docs, and enterprise references.

## Verified Claims

### 1. macOS layer-hosting mode (`darwin/window.go`) ✅
Apple docs confirm: `setLayer:` **before** `setWantsLayer:YES` = layer-hosting mode (app owns layer). Reversed order = layer-backed mode (AppKit manages layer). On macOS 15+ this causes Metal content to composite above title bar. Correct fix.

### 2. null-attach instead of destroy (`libwayland_csd.go`) ✅
`wl_surface.attach(null, 0, 0)` + commit in sync mode applies atomically with parent commit. Destroying before parent commit discards pending null state → one-frame artifact on GNOME. This matches the Wayland sync subsurface spec: pending state applies on parent commit, but only if the surface exists at commit time.

### 3. KDE SSD detection (`libwayland_xdg.go`) ✅
`zxdg_toplevel_decoration_v1.configure` event listener properly registered. `DecorationMode()` returns 0/1/2. `needCSD` logic covers both "protocol not available" and "compositor rejected SERVER_SIDE". Clean approach.

### 4. Forced csdPendingResize on state change (`libwayland_input.go`) ✅
Solves maximize→fullscreen at same resolution (e.g., 1920×1080 maximized → 1920×1080 fullscreen). Without this, `OnConfigure` sees no size delta → skips `SetPendingCSDResize` → title bar stays visible in fullscreen.

### 5. csdPendingRepaint reset timing (`libwayland_xdg.go`) ✅
Moving reset **after** `ResizeCSD` ensures the `csdPendingRepaint=true` signal from `inputToplevelConfigureCb` reaches `ResizeCSD`'s early-return guard. Previously clearing before the call dropped the signal.

### 6. Focused state sync (`libwayland_input.go`) ✅
`h.csdState.Focused = activated` — previously only set in `OnConfigure` callback. Now synchronized directly in configure parsing, ensuring title bar repaint on focus change.

## Architecture Assessment

The null-attach pattern is superior to our original destroy approach:
- **Atomic**: pending state applied with parent commit (no intermediate frames)
- **No recreate cost**: surface+subsurface stay alive, only SHM buffers released
- **Protocol-correct**: Wayland sync subsurface semantics

The decoration mode listener fills a real gap — without it, KDE Plasma users could get double decorations (SSD assumed but compositor gave CLIENT_SIDE).

## CI Status

10/10 checks pass. No go.mod changes. Clean merge.

LGTM 🎯

PR #314, opened 2026-06-15
fix/306-windows : Add scale windows

lkmavi:
#Changes

platform_windows.go

- New constants: wmGetDpiScaledSize (0x02E4), monitorDefaultToPrimary, mdtEffectiveDpi
- New procs: MonitorFromPoint, AdjustWindowRectExForDpi (Win10 1607+), AdjustWindowRectEx (fallback), GetDpiForMonitor (shcore.dll, Win8.1+)
- monitorDpiFromPoint() — queries the monitor's DPI before creating the HWND; Returns 96 as a fallback on older OSes
- adjustWindowRectForDpi() — Enables window frame sizing with auto-fallback to AdjustWindowRectEx
- createWindowWin32() — 4-step SDL3 hybrid: guess DPI → pre-scale → CreateWindow → post-verify + SetWindowPos
- wndProc WM_GETDPISCALEDSIZE — Returns the exact outer-size when changing monitors (SDL3/GLFW pattern)
- windowsPlatform.ScaleFactor() — Now works before window creation (PlatScaleProvider)

scale_windows.go

- SystemScaleFactor() via monitorDpiFromPoint — similar to macOS scale_darwin.go

#Demo

https://github.com/user-attachments/assets/56fffe20-1689-4e8e-b85f-e49230568f78

PR #313, opened 2026-06-15
fix/306-macos : Fix ScaleFactor() macos

lkmavi:
# CHANGES
- app.ScaleFactor() — three-tier resolution (Flutter/GLFW pattern): platWindow → PlatScaleProvider → platform.SystemScaleFactor(). Previously returned 1.0 when called before Run().
- Added PlatScaleProvider optional interface to platform.go for managers that can report DPI scale before a window is created.
- darwinPlatform implements PlatScaleProvider via [NSWindow backingScaleFactor] falling back to [NSScreen mainScreen].backingScaleFactor.
- Added darwin.MainScreenScaleFactor() — queries [NSScreen mainScreen].backingScaleFactor directly, safe to call before NSApplication init (mirrors Flutter/GLFW approach).
- Added platform.SystemScaleFactor() with darwin and non-darwin (1.0) build variants.
- Added examples/hidpi/main.go — demo showing logical vs physical texture resolution on Retina.

PR #316, opened 2026-06-15
Fix/306 macos

lkmavi:
# Changes
**RC-1** (physical size is not tracked when switching between monitors)
- darwinWindow now has physW/physH fields (physical pixels, zero-initialized)
- pollEvents() and checkResize() now compare physical sizes, not just logical ones — during the Retina↔1× transition, where 
logical dim values ​​do not change, EventResize is still emitted
- Zero-initialization ensures that the first pollEvents() always emits EventResize with the actual physical size
(solves the RC-2 starting discrepancy)

**RC-2** (initial scale mismatch)
- Solved via zero-init physW/physH: on first poll, 0 != actual → EventResize → the application can reread
ScaleFactor() and reinitialize the fonts

**RC-3** (no delegate for screen change)
- darwin.Window received screenChangedCh chan struct{}{1} + accessor ScreenChangedCh()
- windowDidChangeScreen is registered in registerWindowDelegateClass(): IMP: signals the channel + calls WakeEventLoop()
to unblock WaitEventsWithHandler
- pollEvents() drains the channel before checking the size (reset for the next transition)
- WakeEventLoop() is a new package-level function in application.go, safe to call from an ObjC callback

### DIFF
--- a/examples/hidpi/main.go
+++ b/examples/hidpi/main.go
@@ -38,6 +38,8 @@ func main() {
 		scale := dc.ScaleFactor()
 		pw, ph := dc.FramebufferSize()
 
+		sizeChanged := pw > 0 && ph > 0 && (pw != lastPW || ph != lastPH)
+		if sizeChanged {
 			if texLow != nil {
 				texLow.Destroy()
 			}
--- a/examples/hidpi/main.go
+++ b/examples/hidpi/main.go
@@ -38,8 +38,6 @@ func main() {
 		scale := dc.ScaleFactor()
 		pw, ph := dc.FramebufferSize()
 
-		sizeChanged := pw > 0 && ph > 0 && (pw != lastPW || ph != lastPH)
-		if sizeChanged {
 			if texLow != nil {
 				texLow.Destroy()
 			}
--- a/internal/platform/darwin/application.go
+++ b/internal/platform/darwin/application.go
@@ -358,6 +358,14 @@ func createApplicationDefinedEvent() ID {
 	return ID(result)
 }
 
+// WakeEventLoop posts an empty event to unblock WaitEventsWithHandler.
+// Safe to call from any goroutine or ObjC delegate callback.
+func WakeEventLoop() {
+	if app != nil {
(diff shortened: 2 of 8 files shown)


kolkov (approved):
Enterprise code review completed. Excellent macOS HiDPI fix — solves the Retina↔1× multi-monitor transition that was the root cause of ui#130, ui#129, and f4#119.

## New macOS HiDPI Changes ✅

### 1. Physical size tracking (`platform_darwin.go`) ✅
`physW`/`physH` fields on `darwinWindow` — zero-initialized so first `pollEvents` always emits `EventResize` with actual physical size. Both `pollEvents()` and `checkResize()` now compare logical AND physical dimensions. This catches the critical case: Retina↔1× transition where logical size stays 800×600 but physical changes 1600×1200 ↔ 800×600.

### 2. `windowDidChangeScreen:` delegate ✅
`NSWindowDidChangeScreenNotification` equivalent via delegate method — fires when window moves to display with different backing scale. Channel-based signaling (`screenChangedCh`, cap 1) with coalescing. `WakeEventLoop()` posts empty NSEvent to unblock `WaitEventsWithHandler`. This is the correct macOS pattern — Qt6 uses `QCocoaWindow::windowDidChangeScreen:` for the same purpose.

### 3. Tests ✅
- `TestWindowDelegateRespondsToWindowDidChangeScreen` — verifies delegate registration
- `TestWindowScreenChangedCh` — channel validity and initial empty state

## Needs Rebase

PR #316 is based on main before PR #314 (Windows HiDPI) was merged. Two conflicts:
1. `scale.go` has `!darwin` tag — should be `!darwin && !windows` (PR #314 added Windows)
2. PR includes all #313 changes (already merged) — `app.go`, `PlatScaleProvider`, `scale_darwin.go`, `examples/hidpi/`

Please rebase on current main to resolve. The macOS-specific changes (physical tracking, windowDidChangeScreen, tests) have no overlap with #314.

## CI

10/10 pass on current base. Will need re-run after rebase.

LGTM pending rebase 🎯

PR #335, opened 2026-06-24
feat/window-background : fix macOS live resize + API polish (SetTitle…

lkmavi:
## macOS live resize                                                                                                      
                                                                                                                            
  - `CAMetalLayer.contentsGravity = "resize"` — existing content scales to                                                  
    fill newly-exposed region instead of showing a blank gap (matches                                                       
    Chrome/Safari behaviour)                                                                                                
  - Eager surface reconfigure in `beginFrame` when physical size changes,                                                   
    before `nextDrawable` acquire                                                                                           
                                                                                                                            
  ## API additions                                                                                                          
                                                                                                                            
  - `App.SetTitle(s)` / `App.Title() string` — runtime window title;                                                        
    fixes no-op stubs on macOS, X11, Wayland                                                                                
  - `App.OnFocus(fn)` / `App.HasFocus() bool` — focus callback and query                                                    
    (events already flowed through platform, just not exposed on App)                                                       
  - `NewSeparator() MenuItem` / `(*Menu).AddSeparator() *Menu`                                                              
  - `ErrWindowClosed`, `ErrDeviceLost` error sentinels                                                                      
  - `AppLifecycle.String()` default: `"AppLifecycle(99)"` per Go convention                                                 
    (`syscall.Signal`, `go/token.Token`)                                                                                    
  - `Config.WithResizable(bool)`


kolkov (dismissed):
Thorough, well-structured PR. Validated against enterprise references (winit, SDL3, Flutter, GLFW, Qt6) — all changes align with industry patterns.

**macOS resize fix:** `contentsGravity = "resize"` is the Chrome/Safari approach — stretching last frame beats blank gap. Eager reconfigure in `beginFrame` (size mismatch instead of ScaleChanged-only) eliminates the 1-frame lag. Both correct.

**SetTitle:** macOS one-liner wiring (underlying `darwin.Window.SetTitle` was already implemented), X11 `SetWindowTitle` sets both `WM_NAME` + `_NET_WM_NAME`, Wayland wires through `libwl.SetTitle` + Flush. All platforms verified.

**OnFocus:** Good that lazy init happens before `a.eventSource.OnFocus(fn)` — without it, callback registered before `Run()` would be silently lost. `HasFocus` via `a.focused` field updated in `classifyEvent` is thread-safe (main thread only).

Minor style: `_ = a.EventSource()` looks like error suppression (`_ = err` pattern). Since `EventSource()` returns an interface (not error), simpler to write:
```go
a.EventSource() // ensure eventSource is initialized
```
Go doesn't require `_ =` for non-error return values.

**Tests:** 7 new tests cover chaining, immutability (Config not mutated), callback dispatch sequence `[true, false, true]`, separator factory + multi-item chain. Good coverage.

**One minor request:** `SetContentsGravity` doc comment in `darwin/surface.go:153-154` still says `Use "topLeft" (kCAGravityTopLeft) to prevent stretching during resize` — stale after the switch to `"resize"`. Could you update it to reflect the new gravity choice?

Everything else LGTM.

kolkov (approved):
LGTM — both review comments addressed (doc comment + EventSource style). All CI green.

PR #342, opened 2026-06-24
fix: recover surface on ErrSurfaceOutdated in present() + stop X11 resize flicker

samyfodil:
## Summary

Two related fixes for window resize / DPI / monitor changes on the surface-present path.

### 1. Present path didn't recover from `ErrSurfaceOutdated`

The acquire and present paths handled the **identical** `wgpu.ErrSurfaceOutdated` differently:

- **Acquire** (`recoverFromAcquireError`): reconfigures the surface and logs at `Debug` — correct. Outdated is an expected, recoverable event on resize/DPI/monitor change.
- **Present** (`(*RenderTarget).present()`): just did `slog.Error("PRESENT ERROR", ...)` and **nothing else** — no reconfigure. It happened to self-heal on the next frame's acquire, but it logged a scary ERROR for a routine event and dropped the current frame.

`present()` now mirrors the acquire path: on `ErrSurfaceOutdated` it reconfigures the swapchain and logs at `Debug` (marking `SurfaceLost` only if reconfigure fails), reserving `ERROR` for genuinely unexpected present errors. It then signals the frame loop to re-render the frame once at the new size so it isn't dropped to a blank frame.

### 2. X11 black flicker on resize

The X11 window was created with `CWBackPixel = BlackPixel`, so the X server filled newly-exposed regions with black on every resize before the GPU backend repainted them — visible as black flicker while dragging/resizing. Switched to background pixmap `None` (the X11 default, and what GLFW/SDL/Chromium use): the server leaves existing pixels untouched on resize, so old content persists until the swapchain presents the next frame.

The two are complementary — the renderer reconfigures and repaints, while the X server no longer blanks the window in the gap.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- Manually verified on X11 + NVIDIA (pure-Go Vulkan backend, `CGO_ENABLED=0`): dragging and resizing the window between monitors no longer logs `PRESENT ERROR ... surface outdated` (now `Debug`), no longer flashes black, and rendering recovers cleanly.

## Trade-off

With background pixmap `None`, the window shows undefined pixels for the brief moment between first map and first frame instead of black. In practice the first frame paints immediately so this is invisible, but it's the documented consequence of the standard anti-flicker setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,12 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Surface-outdated on present** (@samyfodil, #NNN) — `present()` now reconfigures the
+- **Surface-outdated on present** (@samyfodil, #342) — `present()` now reconfigures the
   surface on `wgpu.ErrSurfaceOutdated` (resize/DPI/monitor change) and logs at Debug,
   matching the acquire path (`recoverFromAcquireError`), instead of logging a spurious
   ERROR with no recovery. After reconfiguring, the frame is re-rendered once at the new
   size so resize no longer drops to a blank frame.
-- **X11 black flicker on resize** (@samyfodil, #NNN) — the window was created with a black
+- **X11 black flicker on resize** (@samyfodil, #342) — the window was created with a black
   `CWBackPixel`, so the X server painted newly-exposed areas black on every resize before
   the GPU repainted. Use background pixmap `None` instead (X11 default; GLFW/SDL/Chromium
   pattern) so old pixels persist until the swapchain presents the new frame.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   surface on `wgpu.ErrSurfaceOutdated` (resize/DPI/monitor change) and logs at Debug,
   matching the acquire path (`recoverFromAcquireError`), instead of logging a spurious
   ERROR with no recovery. After reconfiguring, the frame is re-rendered once at the new
-  size so resize no longer flashes a dropped (black) frame.
+  size so resize no longer drops to a blank frame.
+- **X11 black flicker on resize** (@samyfodil, #NNN) — the window was created with a black
+  `CWBackPixel`, so the X server painted newly-exposed areas black on every resize before
+  the GPU repainted. Use background pixmap `None` instead (X11 default; GLFW/SDL/Chromium
+  pattern) so old pixels persist until the swapchain presents the new frame.
 
 ## [0.42.5] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.42.6] - 2026-06-24
+
+### Fixed
+
+- **Surface-outdated on present** (@samyfodil, #NNN) — `present()` now reconfigures the
+  surface on `wgpu.ErrSurfaceOutdated` (resize/DPI/monitor change) and logs at Debug,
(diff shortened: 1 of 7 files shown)


kolkov (commented):
Strong second contribution! Both fixes are enterprise-pattern-validated.

**present() recovery:** Correct mirror of `recoverFromAcquireError` — `ErrSurfaceOutdated` is expected (resize/DPI/monitor), not an error. Debug log + reconfigure + re-render signal. Single retry in `renderFrameMultiThread` prevents infinite loop. Clean.

**X11 CWBackPixmap=None:** Verified against enterprise references:
- SDL3: `xattr.background_pixmap = None` (SDL_x11window.c:662)
- GTK4: `xattributes->background_pixmap = None` (gdksurface-x11.c:4918)
- GLFW: no CWBackPixel/CWBackPixmap at all (no background = X server default = None)

Correct approach — old pixels persist until swapchain presents.

**CHANGELOG:** Good, correct PATCH classification.

**One fix needed before merge:**

`EndFrame()` (renderer.go:541) calls `r.endFrameForSurface(r.primary)` which now returns `bool`, but `EndFrame` ignores the return. Single-window apps using `EndFrame` path don't get re-render-on-outdated — the frame drops to black until the next frame. Should mirror the `renderFrameMultiThread` pattern:

```go
func (r *Renderer) EndFrame() {
    if !r.primary.frameStarted {
        r.pollSubmissions()
        return
    }
    if r.endFrameForSurface(r.primary) {
        r.primary.prepareLazyAcquire()
        // re-render with the same callback that produced the original frame
        // For EndFrame users this may not be feasible — at minimum, 
        // re-acquire + present so the reconfigured surface isn't blank
    }
    r.pollSubmissions()
}
```

If full re-render isn't possible in `EndFrame` (no draw callback reference), at least handling the bool to avoid inconsistency. The next frame will recover either way, but dropping silently is not enterprise behavior.

Everything else LGTM — approve after this fix.

kolkov (changes requested):
Requesting change: EndFrame() (renderer.go:541) ignores the new bool return from endFrameForSurface — single-window apps using EndFrame path don't get re-render-on-outdated. Please handle the return value consistently with renderFrameMultiThread. See detailed review in previous comment.

samyfodil:
@kolkov  Thanks.

`EndFrame` has no retained draw callback or recorded draw sequence to replay. By that point, the caller has already issued the draw operations, so the renderer cannot produce a replacement frame itself.

Calling `prepareLazyAcquire()` there would only reset lazy-acquire state. It is useful immediately before running drawing code, but `EndFrame` has no drawing step after it. Likewise, reacquiring and presenting without rendering would present an unrendered swapchain image with unspecified contents, not reliably restore the previous frame.

The bool remains handled where it can be acted on: `renderFrameMultiThread` retains `frame.onDraw` and reruns it once after `endFrameForSurface` reconfigures an outdated surface. I also changed that retry to use the live `ws.platWindow.ScaleFactor()` rather than the captured frame scale. If the retry also reconfigures during `present`, it requests another redraw for render-on-demand mode.

For X11, `CWBackPixmap=None` means the server does not repaint exposed regions with a background. `Expose` now reaches `RequestRedraw`, so those regions get rendered again.

kolkov (approved):
Thanks for pushing back on the EndFrame point — you were right. I checked wgpu-rs, GLFW, Qt6, GTK4, and SDL3: **none** of them re-render on present-outdated. The universal pattern is reconfigure + drop frame + next iteration handles it. Your EndFrame comment is the correct enterprise approach.

The re-render in `renderFrameMultiThread` actually goes **beyond** what any reference does — a nice enhancement that avoids one dropped frame. Live `ScaleFactor()` + single retry + `RequestRedraw` fallback is clean.

X11 `CWBackPixmap=None` + Expose→RequestRedraw matches SDL3/GLFW/Chromium exactly.

**One minor (non-blocking):** `present()` handles `ErrSurfaceOutdated` but not `ErrSurfaceLost`. `recoverFromAcquireError` handles both (line 523: sets `ws.state = SurfaceLost`). For full symmetry, present() could add:

```go
if errors.Is(err, wgpu.ErrSurfaceLost) {
    slog.Error("gogpu: surface lost on present", "err", err)
    ws.state = SurfaceLost
    return false
}
```

Not blocking — next frame's acquire would catch it. But symmetric error handling is cleaner.

Everything else LGTM. Welcome to the project as a regular contributor, @samyfodil — two strong PRs.

PR #347, opened 2026-06-28
feat/add-golden-tests : Add golden tests + change renderer

lkmavi:
@kolkov Added tests for window examples; run them on Windows and save reference PNGs for verification on other platforms. 

```bash
 go test -run TestGolden_Examples -args -update-golden 
```

### DIFF
--- a/golden_test.go
+++ b/golden_test.go
@@ -1,0 +1,500 @@
+package gogpu
+
+// Golden tests for example rendering output.
+//
+// These tests compare GPU-rendered frames against reference PNG images stored
+// in testdata/golden/examples/. Reference images are generated once on a
+// machine with a working GPU (e.g., Windows + discrete GPU) and committed to
+// the repo. Future runs on other OSes/GPUs compare against those references
+// with a per-pixel threshold to allow for minor driver/backend differences.
+//
+// Workflow:
+//
+//  1. On a reference machine (Windows/GPU): go test -run TestGolden -args -update-golden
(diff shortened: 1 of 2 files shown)


kolkov (changes requested):
Great addition! `RenderToImage` + golden tests is exactly the infrastructure we need for CI-level visual regression testing.

**RenderToImage architecture:** Clean headless GPU rendering — offscreen texture → synthetic RenderTarget → draw callback → CopyTextureToBuffer → MapAsync → readback. BGRA↔RGBA swap, 256-byte row alignment, proper defer chain. Well-extracted `renderToImageReadback` for readability.

**Golden tests:** 9 example scenes with pixel-by-pixel threshold comparison, `-update-golden` flag, graceful skip when no GPU available. Clear documentation on which examples are excluded (and why).

**Two minor items:**

1. `view` created by `CreateTextureView` (renderer.go) is never released — `offscreen` has `defer Release()`, but the view is a separate GPU resource. Add `defer view.Release()` after creation.

2. `RenderToImage` lives on the unexported `Renderer` type — currently only reachable from package-internal tests. If we want this for public headless rendering (server-side image generation, CLI tools), consider exposing it on `App` or `Context` in a follow-up.

Everything else LGTM — CI 10/10 green including codecov.

kolkov (approved):
LGTM — view.Release fix + TODO for public API. 10/10 CI green.

PR #352, opened 2026-06-29
feat/header-positioning-and-resize-fix :   Add header alignment suppo…

lkmavi:
## Summary                                                                                                                
  - Add HeaderAlignment interface for cross-platform title bar alignment control                                            
  - macOS: Implement custom NSTextField injection for center and right alignment                                            
  - Linux/Wayland: Add CSD title alignment support via DefaultCSDPainter                                                    
  - Windows/X11: Stub implementations (platform-controlled title positioning)                                               
  - Update Config with HeaderAlignment field and builder method 
  
 ## Changes                                                                                                                
  - New window_header.go with HeaderAligner interface and applyHeaderAlignment helper                                       
  - darwin/window.go: injectTitleTextField logic for custom title positioning                                               
  - Wayland CSD support for title alignment state and rendering                                                             
  - Secondary window support for header alignment on all platforms  
 
 ## Test plan                                                                                                              
  - [x] macOS: Verify center/left/right alignment rendering correctly                                                       
  - [x] Wayland: Test CSD title alignment with DefaultCSDPainter                                                            
  - [x] Secondary windows: Confirm alignment applies to new windows                                                         
  - [x] Cross-platform: Verify stubs don't break Windows/X11 
  
  
Снимок экрана 2026-06-29 в 23 51 17
Снимок экрана 2026-06-29 в 23 52 01

### DIFF
--- a/app.go
+++ b/app.go
@@ -492,6 +492,8 @@ func (a *App) initPlatform() (platform.PlatformWindow, error) {
 		a.applyHitTestCallback()
 	}
 
+	applyHeaderAlignment(platWindow, a.config.HeaderAlignment)
+
 	// Ensure input subsystems exist. Both EventSource() and Input() use
 	// lazy init so callers can register callbacks before Run(). We must
 	// NOT overwrite instances that were already created — UI frameworks
@@ -514,6 +516,30 @@ func (a *App) initPlatform() (platform.PlatformWindow, error) {
 	// Future: An independent render thread running on its own schedule
 	// would eliminate this callback entirely. See ROADMAP.md for details.
 	a.platWindow.SetModalFrameCallback(a.modalFrameTick)
+
+	// Enable rendering during macOS live resize.
+	//
+	// On macOS, AppKit runs a modal tracking loop during window resize
+	// (inside sendEvent:) that blocks our outer event loop. Without this hook
+	// the old frame sits in the CAMetalLayer at the old size and macOS stretches
+	// it to fill the new window dimensions, producing visible distortion.
+	//
+	// windowDidResize: fires on every resize step from within AppKit's tracking
+	// loop. We call renderFrameMultiThread from the delegate callback goroutine;
+	// the render goroutine is idle while AppKit's modal loop holds the main
+	// goroutine, so RunOnRenderThreadVoid proceeds without deadlock.
+	if lr, ok := platWindow.(platform.LiveResizeRenderer); ok {
+		lr.SetLiveResizeHook(func() {
+			if a.renderLoop == nil {
+				return
+			}
+			physW, physH := platWindow.PhysicalSize()
+			if physW > 0 && physH > 0 {
+				a.renderLoop.RequestResize(uint32(physW), uint32(physH)) //nolint:gosec // G115: validated positive
+			}
+			a.renderFrameMultiThread()
+		})
+	}
 
 	return platWindow, nil
 }
@@ -562,8 +588,9 @@ func (a *App) processEventsMultiThread() {
 	}
 
 	// Queue primary window resize for render thread (deferred pattern).
-	// Don't apply resize during modal resize loop (Windows).
-	if lastResize != nil && a.platWindow != nil && !a.platWindow.InSizeMove() {
+	// RequestResize is an atomic coalesced store so rapid live-resize events
+	// (macOS) collapse to one configure() call per render frame.
+	if lastResize != nil && a.platWindow != nil {
 		// Queue PHYSICAL size for render thread (GPU surface reconfiguration)
 		physW, physH := lastResize.PhysicalWidth, lastResize.PhysicalHeight
 		if physW > 0 && physH > 0 {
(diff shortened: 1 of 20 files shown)


kolkov (commented):
## Code Review

Great work! Two solid features — header alignment API and macOS live resize rendering. The CSD global handle fix is a critical multi-window correctness improvement.

### What's good

- **Optional interface pattern** (`HeaderAligner`, `LiveResizeRenderer`) — matches our established `PlatformWindowCloser` pattern with type assertions, clean
- **CSD pointer callbacks**: migrating from global `csdCallbackHandle` to per-instance `data unsafe.Pointer` fixes a real multi-window bug where all CSD windows shared one handle
- **macOS WindowID**: adding `WindowID: w.id` to all 6 event types (resize, focus, pointer, scroll, key, char) is critical for multi-window event routing — previously these events had no window association
- **`atomic.Bool` for `inLiveResize`** — correct cross-thread pattern, zero non-atomic access
- **`resizeSecondaryWindowsDuringModal()`** — proper Win32 modal loop handling
- **406 lines of tests** with good coverage of nil platform, round-trip, interface compliance, per-window independence
- Well-documented example

### Issues to fix

**1. `CSDState.TitleAlignment` comment is wrong**

`internal/platform/wayland/csd_painter.go` — the field comment says:
```go
TitleAlignment int // 0 = left (default), 1 = center, 2 = right
```
But `PaintTitleBar` treats `case 0` as Center, `case 2` as Right, `default` (1) as Left — matching the public constants (`HeaderAlignCenter=0, HeaderAlignLeft=1, HeaderAlignRight=2`). The **code is correct**, the comment is backwards. Fix:
```go
TitleAlignment int // 0 = center (default), 1 = left, 2 = right — matches gogpu.HeaderAlignment constants
```

**2. Missing blank line** between functions

`internal/platform/platform_darwin.go:334-335`:
```go
}
func (dw *darwinPlatformWindow) SetTitle(title string) {
```
Go convention — blank line between function declarations. Every other adjacent pair in this file has one.

### Minor / follow-up

- **`createSecondaryConn` complexity**: the `//nolint:gocognit,maintidx` is honest, but the function grew from 138 to 218 lines (+80) in this PR. Consider extracting `setupSecondaryCSD()` helper in a follow-up.
- **No CSD painter test**: the new title alignment math in `PaintTitleBar` (center boundary, right fallback, button overlap avoidance) has no unit test. Worth adding `csd_painter_test.go` coverage for the 3 alignment modes.
- **Commit messages**: squash merge with `feat: header alignment API + macOS live resize rendering`

Solid PR overall. Fix the comment + blank line and it's ready to merge.

kolkov (commented):
## Deep Review — Thread Safety & Regression Analysis

After the initial review fixes (all addressed — thank you!), we ran a deeper analysis focusing on thread safety and potential regressions across all platforms. Here are the findings.

### ✅ Confirmed Safe

**InSizeMove guard removal (Windows)** — the guard in `processEventsMultiThread` was redundant. During Win32 modal resize, `DefWindowProc` blocks the main thread in its own message loop, so `processEventsMultiThread` never runs. Two independent suppression layers (WM_SIZE event gating + OS modal loop blocking) ensure no double-resize. Safe to remove.

**Wayland CSD callback ABI** — changing `data uintptr` → `data unsafe.Pointer` in pointer callbacks is ABI-safe. Same register, same bits. goffi explicitly supports `unsafe.Pointer` via reflection dispatch. The `*LibwaylandHandle` is rooted in platform structs, so no GC risk.

### ⚠️ One fix requested before merge

**Live resize hook use-after-free on window close**

In `NewWindow()`, the live resize hook captures `capturedWS` (the `*RenderTarget`). When a secondary window is closed, `closeSecondaryWindow()` destroys the surface but does **not** clear the hook. If AppKit fires a stale `windowDidResize:` after close but before deallocation, the hook would call `resize()` on a destroyed surface.

The delegate dissociation in `Destroy()` mitigates this in practice, but the proper fix is to nil the hook before destroying the surface. In `closeSecondaryWindow()` in `window_manager.go`, before `w.surface.destroy()`:

```go
if lr, ok := w.platWindow.(platform.LiveResizeRenderer); ok {
    lr.SetLiveResizeHook(nil)
}
```

### 📝 Known limitation (follow-up, not a blocker)

**macOS WaitEvents key-window handler routing** — `WaitEvents()` now uses the key (focused) window's `handleEvent`. If an NSEvent belongs to a non-key window (e.g., mouse move on window B while window A is focused), the handler-side state (pointer coords, queued events) is applied to the key window with wrong coordinates. `[NSApp sendEvent:]` still routes to the correct NSWindow, but the gogpu event queue may contain misattributed events.

This only affects multi-window apps (single-window is identical to main branch). We'll track this as a follow-up for multi-window Phase 2 — the correct fix is to extract the target window from `[NSEvent window]` and route to the matching handler.

### Summary

Great PR overall — the CSD per-instance fix, WindowID routing, and live resize rendering are significant multi-window infrastructure improvements. Just the one `SetLiveResizeHook(nil)` fix needed before merge.

lkmavi:
@kolkov Clear SetLiveResizeHook before surface destroy to prevent use-after-free when AppKit fires windowDidResize: on a closing secondary window. Route each NSEvent to its owning NSWindow via [NSEvent window] so pointer coordinates and WindowIDs are attributed to the correct window in  multi-window apps; fall back to key window for window-less events.

kolkov (approved):
All review issues addressed. CSD per-instance fix, WindowID routing, live resize rendering, and the WaitEvents event-window routing fix are solid. LGTM.

PR #353, opened 2026-06-30
feat(x11): wire _NET_WM_ICON for taskbar/decoration icon support

samyfodil:
## Summary

- Adds `Config.WithIcon(image.Image)` to set a window icon on X11 via the EWMH `_NET_WM_ICON` property
- Converts `image.Image` to the CARDINAL/32 ARGB wire format (non-premultiplied `0xAARRGGBB`) expected by EWMH, handling un-premultiplication from Go's `color.RGBA()` output
- No-op on Windows/macOS where icons come from app resources

## Changes

- **`x11/atoms.go`**: intern `_NET_WM_ICON` in `StandardAtoms` at connection time
- **`x11/window.go`**: add `SetNetWMIcon(*Connection, ResourceID, *StandardAtoms, image.Image)` — encodes image to EWMH wire format and calls `ChangeProperty`
- **`x11/platform.go`**: add `Icon image.Image` to x11 `Config`; call `SetNetWMIcon` during window init alongside PID/window-type (non-fatal)
- **`platform/platform.go`**, **`platform_linux.go`**, **`config.go`**, **`app.go`**: thread `Icon image.Image` through every config layer

## Test plan

- [ ] `go test ./internal/platform/x11/... -run TestSetNetWMIcon_E2E -v` — connects to a real X server, creates a window, sets the icon, reads back `_NET_WM_ICON` via `GetProperty`, verifies ARGB encoding for four pixel cases (opaque, semi-transparent, transparent, mixed); skips cleanly in CI without `DISPLAY`
- [ ] `go build ./...` — clean build across all platforms

### DIFF
--- a/app.go
+++ b/app.go
@@ -480,6 +480,7 @@ func (a *App) initPlatform() (platform.PlatformWindow, error) {
 		MinHeight:  a.config.MinHeight,
 		MaxWidth:   a.config.MaxWidth,
 		MaxHeight:  a.config.MaxHeight,
+		Icon:       a.config.Icon,
 	})
 	if err != nil {
 		return nil, err
--- a/app.go
+++ b/app.go
@@ -431,12 +431,15 @@ func (a *App) Run() error {
 	return nil
 }
 
+// newPlatformManagerFn creates a platform manager; replaced in tests to inject mocks.
+var newPlatformManagerFn func() platform.PlatformManager = platform.NewManager
+
 // initPlatform initializes the platform manager, applies pending menu,
 // and creates the primary window. All operations must be on the main thread.
 func (a *App) initPlatform() (platform.PlatformWindow, error) {
 	// Initialize platform manager (process-level) — must be on main thread.
 	platform.SetLogger(slogger())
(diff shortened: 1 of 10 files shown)


kolkov (approved):
Nice work @samyfodil — clean feature, and the follow-up commits addressed everything: `applyWindowProperties` extraction fixed the lint, formatting is clean, and the additional tests (`NoOp` early-returns, `PropagatesIcon` through initPlatform, `WithIcon` builder) bring good coverage. CI is fully green including codecov.

The EWMH encoding and un-premultiplication math are correct — we validated the formula across all 65K `(R,A)` combinations, max rounding error is 1/255 per channel which is invisible for icon pixels.

One minor note: the E2E test pixel values use 0/255 for R/G/B channels, which always un-premultiply exactly. A pixel with mid-range RGB (e.g., `NRGBA{R:100, G:50, B:200, A:150}`) would exercise the integer rounding path. Not blocking — just a future hardening idea.

Needs a rebase on current main (v0.43.1) — no conflicts expected. After that, ready to merge.

PR #358, opened 2026-07-07
fix(app): live-resize hook — eliminate IOSurface churn, add transaction-present toggle

lkmavi:
**Summary:**

The previous live-resize hook called `ws.resize()` on every AppKit
`windowDidResize:` callback. On macOS at 60+ Hz drag rate, each call reached
`setDrawableSize:` in the Metal backend, allocating a new CAMetalLayer
IOSurface pool. Core Animation holds the last-presented drawable for ~16 ms
independently of any GPU idle wait, so pools accumulate faster than they are
released — observed peak: **4.43 GB** footprint during a 10-second drag.

**Changes:**

- **`app.go`**: hook no longer calls `ws.resize()`. Instead, `RequestResize`
  queues the physical dimensions for the render thread; swapchain reconfiguration
  happens once per render tick, not once per geometry event. The main thread
  blocks in `renderFrameMultiThread` until the render thread presents,
  satisfying AppKit's expectation for in-tick frame delivery.

- **`LiveResizePhaser`** (new platform interface): `windowWillStartLiveResize:`
  and `windowDidEndLiveResize:` delegates call registered Go hooks.
  `App.initWindow` uses these to toggle `presentsWithTransaction` on the Metal
  surface. Enabled only during the drag — outside that window the render
  goroutine operates on a non-main thread, so transaction-present would defer
  every frame until the next AppKit event (blank window).

- **`LiveResizeRenderer`** (new platform interface): `windowDidResize:` calls
  the registered hook. Checked via type assertion in `App.initWindow` — no
  impact on non-macOS builds.

- **`renderer.go`**: `setTransactionPresent` uses an interface assertion to
  reach `hal.Surface.SetPresentsWithTransaction(bool)` without Metal imports.

**Known limitation:** Zero-stretch *and* zero-leak simultaneously requires
`[drawable present]` dispatched from the main thread, enabling perpetual
`presentsWithTransaction=true`. The current architecture presents from the
render goroutine. `contentsGravity="topLeft"` (set in wgpu-local) eliminates
stretch artifacts; the small black margin during drag is the accepted trade-off.

**Testing:** `go test ./...` passes. Live-resize stress test (60-second drag):
peak footprint under 150 MB, no observable growth trend.

### DIFF
--- a/app.go
+++ b/app.go
@@ -324,9 +324,31 @@ func (a *App) Run() error {
 	}
 
 	platWindow.Show()
+	a.syncInitialSurfaceSize()
 
-	// macOS can report 0×0 PhysicalSize until the first layout pass; configureSurface
-	// during initRenderer may skip. Sync once after Show so the first OnDraw can present.
+	defer a.shutdown(platWindow)
+
+	a.registerPrimaryWindow(platWindow)
+	a.startRunLoop()
+
+	// Main loop with three rendering modes (ADR-023):
+	//   1. IDLE: No activity — block on OS events (0% CPU, <1ms response)
+	//   2. ANIMATING: StartAnimation() — loop active, onUpdate every tick,
+	//      OnDraw ONLY when RequestRedraw() called (demand-driven, <1% GPU)
+	//   3. CONTINUOUS: ContinuousRender=true — OnDraw every VSync (game loop)
+	for a.running && (a.platWindow == nil || !a.platWindow.ShouldClose()) {
+		a.runFrame()
+	}
+
+	a.lifecycle = AppIdle
+	return nil
+}
+
+// syncInitialSurfaceSize resizes the GPU surface once after the window is
+// shown. macOS can report 0×0 PhysicalSize until the first layout pass, so
+// configureSurface during initRenderer may skip; this sync lets the first
+// OnDraw present.
+func (a *App) syncInitialSurfaceSize() {
 	a.renderLoop.RunOnRenderThreadVoid(func() {
 		if a.renderer != nil && a.platWindow != nil && a.renderer.primary != nil {
 			pw, ph := a.platWindow.PhysicalSize()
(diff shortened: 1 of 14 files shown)


kolkov (commented):
@lkmavi — solid macOS Metal work. We validated each pattern against enterprise references (wgpu-rs, Flutter/Impeller, Skia, Gio, winit, SDL3). Here are the findings.

**Validated patterns (good to merge):**

- **NSAutoreleasePool per frame** — all 5 references do this. wgpu-rs is actually more granular (per-operation pools in `surface.rs:198`, `device.rs`, `command.rs`). Flutter has the exact same `NSAutoreleasePool.new + drain` wrapper (`scoped_nsautorelease_pool.cc:16-23`). Your `RunInFramePool` is correct.

- **Transaction present toggle during resize** — wgpu-rs has the mechanism (`surface.rs:167`, `mod.rs:486-508`). Flutter toggles per-frame (`gpu_surface_metal_skia.mm:155`). Your toggle-on-windowWillStartLiveResize/windowDidEndLiveResize is correct — Flutter explicitly warns against leaving `presentsWithTransaction=true` on the raster thread outside a CA transaction context.

- **allowsNextDrawableTimeout=false** — validated by wgpu-rs (`surface.rs:178`) and Gio (`metal_darwin.go:39`). Already in our wgpu HAL (`surface.go:105`), PR documents existing behavior.

- **Double Resize** — redundant but harmless. `Surface.Configure` is no-op when dimensions match. Both calls serve different edge cases (live resize vs 0×0 init).

**One issue to fix before merge:**

`WaitForGPU()` in `endFrameForSurface` (renderer.go ~line 553) runs on **ALL backends**, not just Metal. This is a full device idle wait (`vkDeviceWaitIdle` / DX12 fence drain) on every frame with GPU work — destroys CPU/GPU pipeline overlap on Vulkan and DX12.

Enterprise references do this differently:
- wgpu-rs: `waitUntilScheduled` (not `waitUntilCompleted`) only in Metal transaction present path (`mod.rs:504`)
- Flutter: `waitUntilScheduled` only in Metal transaction path, `addScheduledHandler` (async) otherwise (`surface_mtl.mm:298-307`)
- Our `Queue.Present()` at `queue.go:469` **already calls** `waitUntilScheduled` when `presentsWithTransaction` is true

So the correct sync is already happening inside `Queue.Present()`. The extra `WaitForGPU()` before present is strictly stronger than what any reference does. If the "blank window" you observed was real, the root cause is likely the draw command buffer not being committed before `Present()` creates its separate present command buffer — which should be fixed at the command buffer level, not with a full device wait.

**Fix:** either remove `WaitForGPU()` from `endFrameForSurface`, or guard it to Metal-only + transaction-present-only. The `Queue.Present()` already handles the correct synchronization.

Everything else is enterprise-validated. The Run() refactor, `pendingRedraw`, `InSizeMove` guard restoration, and 700 lines of tests are solid. Needs rebase on v0.44.0.

kolkov (approved):
Enterprise-validated. WaitForGPU removed. LGTM.

Issue #368, opened 2026-07-11
Wayland app exits after wl_display_flush failed: -1


kivutar:
## Summary

On current `main` (tested via the published `v0.44.3` module resolution), a native Wayland application can close immediately after startup because `wl_display_flush` returns `-1` and gogpu treats that as a fatal dispatch error.

The same application stays alive under X11/XWayland on the same machine.

## Environment

- OS/session: KDE Plasma on Wayland
- GPU: Intel Iris Xe
- Backend: Vulkan
- Display scale: 150% fractional scaling
- `WAYLAND_DISPLAY` set for the failing case

## Reproduction

Using goro as a real app using gogpu:

```sh
GOGPU_LOG=debug timeout 4s ./goro \
  --no-ui \
  --vsync=false
```

With native Wayland, the app exits immediately:

```text
ERROR msg="wayland dispatch error — closing window" error="wl_display_flush failed: -1"
warning: queue ... destroyed while proxies still attached:
  wl_data_offer#4278190080 still attached
  wl_callback#49 still attached
EXIT_CODE=0
```

With X11/XWayland forced, the same app keeps running until `timeout` kills it:

```sh
WAYLAND_DISPLAY= timeout 4s ./goro ...
# EXIT_CODE=124
```

## Expected behavior

A transient flush failure should not immediately close the window if it is a non-fatal Wayland socket backpressure case.

## Actual behavior

The app logs `wayland dispatch error — closing window` and exits.

## Suspected cause

`wl_display_flush` returning `-1` is likely being treated as always fatal. If the underlying errno is `EAGAIN`, the client should normally wait for the socket to become writable and retry rather than closing the window.

I can provide a smaller repro if useful; this was observed while validating Wayland HiDPI/fractional-scale behavior in a larger gogpu app.

PR #369, opened 2026-07-11
Fix Wayland fractional scaling

### LINKED ISSUE #166: Support fractional scaling
I'm currently unable to have sharp widget outlines and sharp text on KDE with 150% fractional scaling. 

QT_SCALE_FACTOR=1 gives me blurry rendering.
QT_SCALE_FACTOR=1.5 gives me sharp rendering but too large widgets.

kivutar:
## Summary

Adds native Wayland fractional-scale support using:

- `wp_fractional_scale_manager_v1` to read the compositor preferred scale
- `wp_viewporter` to keep the surface destination in logical coordinates
- per-frame surface resizing when the physical framebuffer size changes
- `wl_output` integer scale as a fallback when fractional scale is unavailable

## Why

On KDE Wayland at 150% scaling, gogpu was rendering the framebuffer at the logical window size and the compositor scaled it up, making apps visibly blurry.

Observed before this fix:

```text
ctx=1504x1003 framebuffer=1504x1003 app=1504x1003 scale_factor=1.000
```

With the fix, the app renders to the physical framebuffer while keeping logical coordinates stable:

```text
ctx=1504x1003 framebuffer=2256x1505 app=1504x1003 scale_factor=1.500 framebuffer_scale=1.500x1.500
```

This matches the behavior observed through XWayland on the same KDE setup, where the framebuffer is already scaled correctly.

## Notes

Runtime validation against current `main` is currently affected by #368 (`wl_display_flush failed: -1` closes native Wayland apps). The fractional-scale path was validated in the same app on the v0.42.7 baseline, then rebased onto current `main` and reviewed to keep the patch scoped to Wayland scale handling.

## Tests

```sh
go test ./...
```

Fixes https://github.com/gogpu/ui/issues/166

### DIFF
--- a/app.go
+++ b/app.go
@@ -19,6 +19,17 @@ type appRenderLoop interface {
 	ConsumePendingResize() (uint32, uint32, bool)
 }
 
+func physicalSizeUint32(width, height int) (uint32, uint32, bool) {
+	if width <= 0 || height <= 0 {
+		return 0, 0, false
+	}
+	maxUint32 := uint64(^uint32(0))
+	if uint64(width) > maxUint32 || uint64(height) > maxUint32 {
+		return 0, 0, false
+	}
+	return uint32(width), uint32(height), true
+}
+
 // App is the main application type.
 // It manages the window, rendering, and application lifecycle.
 //
@@ -612,8 +623,8 @@ func (a *App) setupLiveResizeHook(platWindow platform.PlatformWindow) {
 		// on every Resized event). onResize (user layout callback) still
 		// fires once after the drag via the normal event path.
 		physW, physH := platWindow.PhysicalSize()
-		if physW > 0 && physH > 0 {
-			a.renderLoop.RequestResize(uint32(physW), uint32(physH)) //nolint:gosec // G115: validated positive
+		if w, h, ok := physicalSizeUint32(physW, physH); ok {
+			a.renderLoop.RequestResize(w, h)
 		}
 		a.renderFrameMultiThread()
 	})
@@ -696,8 +707,8 @@ func (a *App) processEventsMultiThread() {
 	if lastResize != nil && a.platWindow != nil && !a.platWindow.InSizeMove() {
 		// Queue PHYSICAL size for render thread (GPU surface reconfiguration)
 		physW, physH := lastResize.PhysicalWidth, lastResize.PhysicalHeight
-		if physW > 0 && physH > 0 {
-			a.renderLoop.RequestResize(uint32(physW), uint32(physH)) //nolint:gosec // G115: validated positive
+		if w, h, ok := physicalSizeUint32(physW, physH); ok {
+			a.renderLoop.RequestResize(w, h)
 		}
 
 		// Call user callback with LOGICAL size (what user expects for layout)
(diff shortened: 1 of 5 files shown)


kolkov (changes requested):
Great work on this, @kivutar — fractional scale support is something we've been wanting for a while, and the overall approach is solid. The protocol interface descriptors, the scale calculation (`/120.0`), and the fallback chain are all correct.

I went through the code carefully and cross-referenced with how winit, SDL3, and GTK4 handle `wp_fractional_scale_v1` + `wp_viewporter`. A few things came up that I'd like to address before merging.

### CI

Two lint errors need fixing:

- **gosec G115** in `app.go:1013` — `uint32(frame.physW)` needs a positivity guard. We use `//nolint:gosec // G115: validated positive` in similar spots (see `app.go:616`).
- **gocyclo** — `renderFrameGPU` went from 20 to 21. Either extract the new block into a small helper or add `//nolint:gocyclo`.

### Protocol destroy before proxy destroy

In `Close()`, the four new objects (`viewport`, `viewporter`, `fractionalScaleObj`, `fractionalScaleMgr`) are cleaned up with `proxyDestroy` only. Wayland protocol requires sending the `destroy` request (opcode 0) first — that tells the compositor to release server-side resources. Without it, the compositor leaks the object.

Our existing code does this for cursor shape manager (`libwayland.go:374`):
```go
h.marshalVoid(h.cursorShapeMgr, 0)
h.proxyDestroy(h.cursorShapeMgr)
```

Same pattern needed for all four objects. SDL3 and GTK4 both send the protocol destroy before freeing the proxy.

### `SetViewportDestination` per frame

The current PR calls `SetViewportDestination` in `PrepareFrame` every frame. All three references I checked avoid this:

- **winit** calls it only in the resize/configure handler
- **SDL3** calls it only when `window_size_changed || buffer_size_changed`
- **GTK4** uses a `viewport_dirty` flag

The call is idempotent so it won't break anything, but it sends a protocol message on every frame unnecessarily. A simple change-detection (`lastViewportW/H`) or dirty flag would match the enterprise pattern and avoid IPC noise.

### Per-window fractional scale

In the Wayland protocol, fractional scale and viewport are per-surface — winit, SDL3, and GTK4 all create separate `wp_fractional_scale_v1` + `wp_viewport` objects for each surface.

Right now, `PrepareFrame` accesses `w.platform.libwl` (the primary window's handle), and `ScaleFactor()` delegates to `w.platform.ScaleFactor()`. This means all windows share the primary window's scale and viewport. For a secondary window at scale > 1.0, `SetViewportDestination` would set the **primary** viewport's destination using the secondary window's logical size — corrupting the primary window's rendering.

Each `secondaryWaylandConn` already has its own `libwl *LibwaylandHandle` with its own fractional scale fields. The proper fix:

1. Call `SetupFractionalScale` in `createSecondaryConn` (same as `initSingleConnection`).
2. `PrepareFrame` should use the window's own `libwl` — primary via `w.platform.libwl`, secondary via `w.secondary.libwl`.
3. `ScaleFactor()` should query per-window fractional scale from the window's own `libwl`, not the shared platform one.

This can be done in this PR or as a tracked follow-up — but the `PrepareFrame` using `w.platform.libwl` unconditionally is a bug that should be fixed before merge.

### Nice-to-have (non-blocking)

- **`set_source(-1, -1, -1, -1)`** — SDL3 explicitly calls this after creating the viewport to select the entire buffer. The Wayland default is "entire buffer" so this isn't strictly needed, but it's more explicit.
- **Cleanup ordering** — the new objects are destroyed between decoration and xdg cleanup. Since they were created after decorations (reverse creation order), they should ideally come first. Not a correctness issue since they're all siblings destroyed before `wl_surface`.

Let me know if you have questions on any of these. Happy to help if you want to pair on the fixes.

kolkov (approved):
All review items addressed. Protocol destroy, viewport cache, per-window fractional scale with `libwayland()` helper, gosec/gocyclo fixes. CI green. Clean work.

PR #381, opened 2026-07-20
fix/linux-window-id-stamping : Add windows ID

lkmavi:


kolkov (changes requested):
Great fix — this was a real bug where per-window callbacks (keyboard, pointer, scroll, char) were silently never invoked on Linux because `WindowID == 0` always missed `getByPlatformID` lookup. Root cause correctly identified, clean refactor with `translateX11Event`, and the Wayland `queueEvent` one-liner is elegant.

X11 multi-window via independent connections is the right architecture — mirrors the existing Wayland pattern exactly.

One issue needs fixing before merge:

### P1: WaitEvents secondary slice race

In `WaitEvents`, the secondaries snapshot is a **slice alias**, not a copy:

```go
p.secondaryMu.RLock()
secs := p.secondaries   // ← alias, shares backing array
p.secondaryMu.RUnlock()
```

`PollEvents` correctly does `copy(secs, p.secondaries)`. `WaitEvents` should match — if `Destroy()` concurrently modifies `p.secondaries`, this races:

```go
p.secondaryMu.RLock()
secs := make([]*secondaryX11Conn, len(p.secondaries))
copy(secs, p.secondaries)
p.secondaryMu.RUnlock()
```

### Minor (non-blocking)

- `WaitEvents` uses 5ms timeout per secondary — with N windows that's 5ms×N added latency per loop. Fine for 1-3 windows, consider a TODO for future optimization if needed.
- `secondaryX11Conn` godoc mentions "geckty/gogpu" — typo?
- Squash merge commit suggestion: `fix(linux): stamp WindowID on all X11/Wayland events + X11 multi-window`

Everything else looks good — tests cover all 12 event types, the "known gap" comment on Wayland is honest and appreciated.

kolkov (approved):
All review items addressed. P1 slice race fixed (make+copy in both PollEvents and WaitEvents). P2 latency improved with budget-split approach (100ms / (N+1), floor 1ms). Typo fixed. Clean work.

PR #395, opened 2026-07-28
fix(x11): prefer server keyboard mapping when remote

### LINKED ISSUE #279: Consider using keytrans instead of relying solely on libxkbcommon for x11 kb layouts
1. libxkbcommon in unreliable under xQuarz
2. You may be running via x forwarding on device not having libxkbcommon, not having libx11, not even having keyboard maps.

keytrans solves both problems by providing 6 different ways of layout processing with best choice auto selection and graceful degradation.

https://GitHub.com/unxed/keytrans

besmpl:
## Summary

Fixes X11 text translation when the client host's keyboard configuration is not authoritative: X forwarding, XQuartz, and XWayland.

Thanks @unxed for pointing to keytrans and its fallback-first design. I evaluated direct integration, but keytrans currently requires a `*xgb.Conn`; gogpu owns a separate pure-Go X11 wire connection, so adopting it directly would add a second X connection/event owner plus overlapping protocol dependencies.

This PR applies the same core principle at gogpu's existing ownership seam instead:

- prefer the connected X server's `GetKeyboardMapping` data for remote `DISPLAY` hosts, XQuartz vendors, and XWayland
- keep xkbcommon as the richer primary translator on native local Xorg
- refresh the server mapping on both core `MappingNotify` and XKB keymap notifications
- refresh the pure-Go mapping even when XKB/xkbcommon is unavailable
- add no dependency and preserve `CGO_ENABLED=0`

This matters for XQuartz in particular because its effective core keymap can be rewritten when the macOS input source changes; a client-host xkbcommon keymap can be valid but wrong.

## Testing

- table tests for local Xorg, remote TCP/SSH `DISPLAY`, XQuartz/Apple vendors, XWayland, and malformed displays
- server-keymap printable/non-printable dispatch tests
- `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go test -c -o /dev/null ./internal/platform/x11`
- `CGO_ENABLED=0 go test -count=1 ./...`
- `CGO_ENABLED=0 go build ./...`

Physical XQuartz and SSH-forwarding validation remains valuable after CI, but the selection and refresh policy is deterministic and covered.

Closes #279

### DIFF
--- a/internal/platform/x11/keyboard_source_test.go
+++ b/internal/platform/x11/keyboard_source_test.go
@@ -1,0 +1,76 @@
+//go:build linux
+
+package x11
+
+import (
+	"testing"
+
+	"github.com/gogpu/gogpu/internal/platform/eventqueue"
+	"github.com/gogpu/gpucontext"
+)
+
+func TestShouldPreferServerKeymap(t *testing.T) {
+	tests := []struct {
+		name     string
+		display  string
+		vendor   string
+		xwayland bool
+		want     bool
+	}{
+		{name: "local Xorg", display: ":0", vendor: "The X.Org Foundation"},
+		{name: "explicit local unix", display: "unix:0", vendor: "The X.Org Foundation"},
+		{name: "SSH forwarding", display: "localhost:10.0", vendor: "The X.Org Foundation", want: true},
+		{name: "remote TCP", display: "workstation.example:0", vendor: "The X.Org Foundation", want: true},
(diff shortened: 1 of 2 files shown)


kolkov (approved):
Enterprise-quality fix. Detection logic solid (8 test cases), server keymap priority for remote/XQuartz/XWayland, backward-compat for local Xorg. APPROVE.

Issue #379, opened 2026-07-19
Draw callback is called 60 times per second. Continuous render is off. Wayland.


gneggo:
I’ve made a program that prints current time in its draw callback. If I understood this correctly, the callback should only run when an event happens, or when I call `app.RequestRedraw`. But my program prints 60 times per second.

I’m using arch linux, wayland via river-classic, go 1.26.5, gogpu v0.44.9.

Code:

```go
package main

import (
	"time"

	"github.com/gogpu/gogpu"
)

func main() {
	app := gogpu.NewApp(gogpu.DefaultConfig().WithContinuousRender(false))

	app.OnDraw(func(ctx *gogpu.Context) {
		ctx.Clear(1, 1, 1, 1)
		println(time.Now().String())
	})

	if err := app.Run(); err != nil {
		panic(err)
	}
}
```

kolkov:
Fixed in v0.45.0. Frame callback is now a GATE (via frameCallbackReady()), not a TRIGGER. Removed unconditional EventExpose synthesis on compositor done event.

PR #411, opened 2026-08-01
app: don't reschedule a frame that drew nothing

samyfodil:
Lazy acquire defers `beginFrame` to the first draw call, so that a callback which issues no draw calls costs nothing — no swapchain acquire, no present. `renderFrameGPU` then treats that same condition, `!ws.frameStarted`, as a **failed** `beginFrame` and calls `RequestRedraw`:

```go
// beginFrame can fail (outdated / not yet configured). Demand-driven mode
// already consumed the invalidation — schedule another frame.
if !ws.frameStarted {
    a.RequestRedraw()
```

For a demand-driven UI those are not the same thing. Such a UI draws nothing whenever nothing changed, which is most frames, so the retry produces another empty frame, which retries again. The loop never blocks and the app renders as fast as the platform allows for as long as it is open.

### What it costs

Found on a terminal emulator built on gogpu/ui. An **idle** window — nothing printing, nothing animating — sat at **27% of a core** and made ~1400 event-loop turns a second, each one an X11 round trip. The widget tree rendered **zero** frames in fourteen seconds, counted in the app's own draw callback, so nothing above the platform layer was asking for any of it. CPU was identical with the window minimised.

Once any redraw happens the app is stuck in that state for the rest of its life, which is why it showed up as a constant idle cost rather than a spike.

This is the same shape as the Wayland frame-callback loop fixed in #379: a readiness signal used as a trigger.

### The fix

`acquireFailed` records that a draw call reached `ensureFrameStarted` and could not have the swapchain. Only that case is owed another frame; a callback that never drew is left alone.

Idle CPU on the terminal that found it: **27% → 0.75%**. Syscalls while idle: **1361/s → 26/s**. Frames still arrive normally on output, input and cursor blink.

### Tests

`TestRenderFrameGPU_NoDrawCallsDoesNotRescheduleForever` is the regression: a callback that draws nothing must not queue another frame.

`TestRenderFrameGPU_UnconfiguredSurface` asserted the old behaviour using a callback that drew nothing, which is exactly the case being fixed. Its name and intent are about an *unconfigured surface*, so it now makes a real draw call that cannot acquire, and still asserts the redraw is queued — the case that genuinely needs a retry.

### DIFF
--- a/app.go
+++ b/app.go
@@ -1101,10 +1101,23 @@ func (a *App) renderFrameGPU(frames []windowFrame) {
 		ctx := newContextForSurface(a.renderer, ws, frame.scale)
 		frame.onDraw(ctx)
 
-		// beginFrame can fail (outdated / not yet configured). Demand-driven mode
-		// already consumed the invalidation — schedule another frame.
+		// No frame started. Two very different reasons:
+		//
+		// The callback issued no draw calls, so lazy acquire never acquired
+		// anything. That is what lazy acquire is for, not a failure — there is
+		// nothing to present and nothing to retry. Asking for another frame
+		// here is a loop with no exit: a demand-driven UI draws nothing when
+		// nothing changed, so the next frame draws nothing either, and the app
+		// spins at whatever rate the platform allows, burning CPU on an idle
+		// window (~27% of a core on X11, with no visible frames at all).
+		//
+		// Or a draw call did ask for the swapchain and could not have it
+		// (surface outdated, not yet configured). Demand-driven mode already
+		// consumed the invalidation, so that one does need another frame.
 		if !ws.frameStarted {
-			a.RequestRedraw()
+			if ws.acquireFailed {
+				a.RequestRedraw()
+			}
 			ws.resetLazyState()
 			a.renderer.currentSurface = nil
 			continue
--- a/app_coverage_test.go
+++ b/app_coverage_test.go
@@ -212,9 +212,37 @@ func TestRenderFrameGPU_NilSurfaces(t *testing.T) {
 }
 
 // TestRenderFrameGPU_UnconfiguredSurface exercises ws!=nil but CanRender()==false
-// with no platWindow set on the RenderTarget.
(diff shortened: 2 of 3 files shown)


kolkov (approved):
Thorough review completed. This is a clean, well-reasoned fix.

### Analysis

**Root cause is correct.** The current code conflates two semantically different reasons for `!ws.frameStarted`:

1. **Callback drew nothing** — lazy acquire working as designed. No swapchain acquire, no present, no retry needed. Demand-driven UIs draw nothing most frames.
2. **Callback drew but swapchain unavailable** — genuine failure, retry warranted.

Treating both as case (2) creates an infinite render loop: empty frame → RequestRedraw → empty frame → RequestRedraw. ~27% CPU on idle X11 — confirmed by the analysis.

### Code validation

- `acquireFailed` set at the right place: `ensureFrameStarted()` line 665, just before `return false` — all draw-call-triggered paths that fail to acquire converge here.
- `acquireFailed` reset in both `prepareLazyAcquire()` and `resetLazyState()` — correct, covers all frame lifecycle paths.
- No edge case: if a second `ensureFrameStarted()` call succeeds after the first failed, `ws.frameStarted=true` → the `if !ws.frameStarted` branch doesn't execute → `acquireFailed` value is irrelevant.
- Test `TestRenderFrameGPU_UnconfiguredSurface` correctly updated: `onDraw` now calls `MarkExternalContent()` (a draw operation) to trigger `ensureFrameStarted` → `acquireFailed=true`. The `renderer: &Renderer{}` addition prevents nil panic on `ws.renderer.device`.
- New test `TestRenderFrameGPU_NoDrawCallsDoesNotRescheduleForever` — validates the fix directly.

### Enterprise quality

- Comments are excellent — explain both cases clearly, include the measured impact (27% CPU)
- Boolean naming `acquireFailed` is precise and self-documenting
- Minimal diff — 3 files, focused change, no unrelated modifications
- Test coverage: existing test updated + new test added

**LGTM. No changes requested.**

Issue #361, opened 2026-07-08
Feature request: per-pixel-alpha layered windows + runtime show/hide/reposition (floating tray-popup UIs)


shaolei:
## Motivation

We're building **DeskCalendar** — an open-source Windows tray calendar popup (a lightweight alternative to the system clock panel, inspired by 360 小清新日历) on top of the GoGPU ecosystem (`gogpu` + `gogpu/ui`). The core UX is a **floating, rounded, semi-transparent card** that pops up next to the taskbar clock when the tray icon is clicked and hides on focus loss.

To achieve this we need three capabilities that, as of `0578472` (v0.44.1), `gogpu` does not yet expose on Windows:

### FR1 — Per-pixel alpha in the swapchain
`renderer.go:293` hardcodes `AlphaMode: gputypes.CompositeAlphaModeOpaque`. Without `CompositeAlphaModePremultiplied` a transparent/rounded root surface cannot composite against the desktop — the rounded corners stay opaque black.

### FR2 — Layered window style (`WS_EX_LAYERED`)
`internal/platform/platform_windows.go` calls `CreateWindowExW` with `dwExStyle = 0` (lines 655 / 662 / 698). For a click-through-transparent floating card we need `WS_EX_LAYERED` (+ `WS_EX_NO_REDIRECTION_BITMAP` on Win10+) so the rounded exterior shows the desktop behind it and DWM can supply the shadow.

### FR3 — Runtime show / hide / reposition
A tray popup must appear at the clock position and hide on blur. `win32Window.Show()` exists internally (`platform_windows.go:947`) but is **not exposed on the public `Window`**; there is **no `Hide()`, `SetPosition(x, y)`, or `SetSize(w, h)`** on Windows at all.

## Precedent: macOS already has this
`internal/platform/darwin/window.go` already implements `Show()` / `Hide()` / `SetSize()` (lines 186 / 200 / 256) on its `Window`. So this request is really **"bring Windows to parity with the existing macOS API" + add the layered/alpha bits** — not a brand-new concept for the project.

## Proposed API (one possible shape — happy to adapt)
- **Window options / `WindowStyle`**: add `Layered bool` and `PerPixelAlpha bool`, driving `WS_EX_LAYERED` and `CompositeAlphaModePremultiplied`.
- **Public `Window` methods**: `Show()`, `Hide()`, `SetPosition(x, y int)`, `SetSize(w, h int)`. The internal `SetWindowPos` plumbing already exists in `platform_windows.go`.

[...]

PR #430, opened 2026-08-06
fix(windows): make DComp per-pixel alpha visible on Win11 + DX12

shaolei:
## Summary

Per-pixel alpha transparency from #361 still did not composite against the desktop on Windows 11 with the DX12 backend. This PR makes it actually visible with three related changes:

1. **Set `WS_EX_NOREDIRECTIONBITMAP` for transparent windows** (`0x00200000`). Without it the DComp swap chain is composed onto the DWM redirection surface: semi-transparent content blends with a fixed background instead of the desktop (matches the wgpu v0.30.35 release note). Verified by A/B: with the style the window shows the desktop through the alpha channel; without it, the same frame blends against white.

2. **Stop calling `DwmEnableBlurBehindWindow` for transparent windows** and remove the now-dead blur-behind helpers. With `WS_EX_NOREDIRECTIONBITMAP`, the blur-behind call paints an opaque blurred backdrop behind the swap chain and hides the desktop (winit skips blur-behind in this mode).

3. **Skip `DwmExtendFrameIntoClientArea` for transparent frameless windows**. The call switches the window onto the DWM glass composition path, which overrides DirectComposition per-pixel alpha — a 50% transparent fill renders as an opaque tint. Frameless transparent windows lose the DWM system shadow (apps can draw their own); `SetWindowPos(swpFrameChanged)` and `updateSize()` are kept so the `WM_NCCALCSIZE` JBR path still removes the title bar.

## Testing

- Windows 11 + DX12, NVIDIA GeForce RTX 4060.
- A 360x480 frameless transparent window with 50% premultiplied alpha shows the desktop through the panel; the same frame is opaque with either of the two removed calls.
- Verified end-to-end in DeskCalendar (tray calendar popup): semi-transparent panel, transparent rounded corners, and low-alpha elements render correctly on both light and dark themes. I will keep testing this in the project and report back here if anything else comes up.

### DIFF
--- a/app.go
+++ b/app.go
@@ -6,6 +6,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/gogpu/gogpu/gpu/types"
 	"github.com/gogpu/gogpu/input"
 	"github.com/gogpu/gogpu/internal/platform"
 	"github.com/gogpu/gogpu/internal/thread"
@@ -569,11 +570,13 @@ func (a *App) initPlatform() (platform.PlatformWindow, error) {
 		Fullscreen:  a.config.Fullscreen,
 		Frameless:   a.config.Frameless,
 		Transparent: a.config.Transparent,
-		MinWidth:    a.config.MinWidth,
-		MinHeight:   a.config.MinHeight,
-		MaxWidth:    a.config.MaxWidth,
-		MaxHeight:   a.config.MaxHeight,
-		Icon:        a.config.Icon,
+		UseDirectComposition: a.config.Transparent &&
+			a.config.GraphicsAPI == types.GraphicsAPIDX12,
+		MinWidth:  a.config.MinWidth,
+		MinHeight: a.config.MinHeight,
+		MaxWidth:  a.config.MaxWidth,
+		MaxHeight: a.config.MaxHeight,
+		Icon:      a.config.Icon,
 	})
 	if err != nil {
 		return nil, err
--- a/app_test.go
+++ b/app_test.go
@@ -4,6 +4,7 @@ import (
 	"image"
 	"testing"
 
+	"github.com/gogpu/gogpu/gpu/types"
 	"github.com/gogpu/gogpu/input"
 	"github.com/gogpu/gogpu/internal/platform"
 	"github.com/gogpu/gpucontext"
@@ -1409,5 +1410,38 @@ func TestInitPlatform_PropagatesIcon(t *testing.T) {
 	}
 	if capturing.got.Icon != img {
 		t.Error("initPlatform: Icon not propagated to platform.Config")
+	}
+}
+
+func TestInitPlatform_PropagatesUseDirectComposition(t *testing.T) {
+	tests := []struct {
+		name        string
+		api         types.GraphicsAPI
+		transparent bool
+		want        bool
+	}{
+		{"auto+transparent", types.GraphicsAPIAuto, true, false},
+		{"vulkan+transparent", types.GraphicsAPIVulkan, true, false},
+		{"dx12+transparent", types.GraphicsAPIDX12, true, true},
+		{"dx12+opaque", types.GraphicsAPIDX12, false, false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			app := NewApp(DefaultConfig().
+				WithGraphicsAPI(tt.api).
+				WithTransparent(tt.transparent))
+
+			capturing := &configCapturingManager{}
+			old := newPlatformManagerFn
+			newPlatformManagerFn = func() platform.PlatformManager { return capturing }
+			defer func() { newPlatformManagerFn = old }()
+
+			if _, err := app.initPlatform(); err != nil {
+				t.Fatalf("initPlatform: %v", err)
+			}
+			if capturing.got.UseDirectComposition != tt.want {
+				t.Errorf("UseDirectComposition = %v, want %v", capturing.got.UseDirectComposition, tt.want)
+			}
+		})
 	}
 }
--- a/app_test.go
+++ b/app_test.go
@@ -4,7 +4,7 @@ import (
 	"image"
 	"testing"
 
-	"github.com/gogpu/gogpu/gpu/types"
+	gpu_types "github.com/gogpu/gogpu/gpu/types"
 	"github.com/gogpu/gogpu/input"
 	"github.com/gogpu/gogpu/internal/platform"
 	"github.com/gogpu/gpucontext"
@@ -1416,14 +1416,14 @@ func TestInitPlatform_PropagatesIcon(t *testing.T) {
 func TestInitPlatform_PropagatesUseDirectComposition(t *testing.T) {
 	tests := []struct {
 		name        string
-		api         types.GraphicsAPI
+		api         gpu_types.GraphicsAPI
 		transparent bool
 		want        bool
 	}{
-		{"auto+transparent", types.GraphicsAPIAuto, true, false},
-		{"vulkan+transparent", types.GraphicsAPIVulkan, true, false},
-		{"dx12+transparent", types.GraphicsAPIDX12, true, true},
-		{"dx12+opaque", types.GraphicsAPIDX12, false, false},
+		{"auto+transparent", gpu_types.GraphicsAPIAuto, true, false},
+		{"vulkan+transparent", gpu_types.GraphicsAPIVulkan, true, false},
+		{"dx12+transparent", gpu_types.GraphicsAPIDX12, true, true},
+		{"dx12+opaque", gpu_types.GraphicsAPIDX12, false, false},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/config.go
+++ b/config.go
@@ -292,6 +292,16 @@ func (c Config) WithFrameless(frameless bool) Config {
 // supported and the platform window is created with its native transparency
 // style. Falls back to an opaque surface when the backend does not support
 // premultiplied alpha.
+//
+// Windows backend matrix (see docs/windows-transparency.md):
+//   - DX12 (explicit): DirectComposition + WS_EX_NOREDIRECTIONBITMAP, full
+//     per-pixel alpha against the desktop.
+//   - Vulkan/GLES: DwmEnableBlurBehindWindow legacy path; per-pixel alpha
+//     works through the DWM redirection surface.
+//   - Software: not currently supported — GDI BitBlt/StretchDIBits does not
+//     preserve alpha. Use a GPU backend or UpdateLayeredWindow.
+//   - Auto: uses the legacy blur-behind path so all backends stay visible;
+//     if wgpu selects DX12, per-pixel alpha may not composite (see docs).
 func (c Config) WithTransparent(transparent bool) Config {
 	c.Transparent = transparent
 	return c
--- a/docs/windows-transparency.md
+++ b/docs/windows-transparency.md
@@ -1,0 +1,70 @@
+# Windows Transparent Windows
+
+## Two presentation paths
+
+On Windows there are two different ways a transparent gogpu window can be
+composited:
+
+### 1. DirectComposition (DX12, explicit)
+
+`WS_EX_NOREDIRECTIONBITMAP` is set at `CreateWindowExW` time and
+`DwmEnableBlurBehindWindow` is **not** called. wgpu creates a
+`CreateSwapChainForComposition` swap chain and presents through a DComp
+visual tree. This is the only path on which a DX12 swap chain's per-pixel
+alpha composites directly against the desktop.
+
+### 2. Legacy DWM blur-behind (Vulkan / GLES / Auto)
+
+The window keeps the DWM redirection surface and
+`DwmEnableBlurBehindWindow(DWM_BB_ENABLE | DWM_BB_BLURREGION)` is called with
+an empty region. The GPU swap chain presents through the HWND and DWM uses the
+alpha channel of the redirection surface. This works for Vulkan (including
+`VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR`, because DWM-level blur handles alpha) and
+GLES.
+
+`WS_EX_NOREDIRECTIONBITMAP` must not be used on this path: Vulkan/GLES/Software
+present through the HWND, and without a redirection surface the window content
+is invisible.
+
+## Verified matrix
+
+Verified on Windows 11 with an NVIDIA GeForce RTX 4060 (driver 591.86) using a
+frameless 50% transparent red panel over a known backdrop:
+
+| API | Window style | 50% alpha result |
+|-----|--------------|------------------|
+| DX12 (explicit) | `WS_EX_NOREDIRECTIONBITMAP` + DComp | per-pixel alpha visible |
+| Vulkan (explicit) | legacy blur-behind | per-pixel alpha visible |
+| GLES (explicit) | legacy blur-behind | per-pixel alpha visible |
+| Software (explicit) | legacy blur-behind | **not supported** — GDI `BitBlt`/`StretchDIBits` (`SRCCOPY`) does not preserve alpha |
+| Auto | legacy blur-behind | visible, but see the DX12 caveat below |
+
+## Known limitations
+
+### Software backend
+
+The software backend presents through GDI `BitBlt`/`StretchDIBits` with
+`SRCCOPY`, which does not carry per-pixel alpha. `DwmEnableBlurBehindWindow`
+alone is not enough. A future implementation should use
+`UpdateLayeredWindow` (or `AlphaBlend` from a DIB section) for transparent
+software windows. Until then, `WithTransparent(true)` + Software should be
+treated as unsupported.
+
+### Auto + DX12
+
+Auto mode deliberately uses the legacy blur-behind path because the backend is
+not known at window creation time and Auto prefers Vulkan. If wgpu selects DX12
+on a Vulkan-less machine, the DX12 backend still creates a DirectComposition
+swap chain whenever `CompositeAlphaModePremultiplied` is requested, but without
+`WS_EX_NOREDIRECTIONBITMAP` the DComp content composites against the opaque
+redirection surface instead of the desktop. The result is a white/opaque tint,
+not per-pixel transparency.
+
+Options:
+
+- Require callers who need true DX12 transparency to set
+  `WithGraphicsAPI(GraphicsAPIDX12)` explicitly.
+- Resolve the backend before window creation (larger refactor).
+- Recreate the window when Auto resolves to DX12 (Godot's approach). This
+  destroys and recreates the HWND, so it can cause a brief visible flash /
+  focus change; doing it before the first `Show` minimizes the visible effect.
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -20,20 +20,21 @@ func NewWindowID() WindowID {
 
 // Config holds platform-agnostic window configuration.
 type Config struct {
-	Title             string
-	Width             int
-	Height            int
-	Resizable         bool
-	Fullscreen        bool
-	Frameless         bool
-	Transparent       bool
-	TabbingMode       int
-	TabbingIdentifier string
-	MinWidth          int // 0 = no minimum constraint
-	MinHeight         int // 0 = no minimum constraint
-	MaxWidth          int // 0 = no maximum constraint
-	MaxHeight         int // 0 = no maximum constraint
-	Icon              image.Image
+	Title                string
+	Width                int
+	Height               int
+	Resizable            bool
+	Fullscreen           bool
+	Frameless            bool
+	Transparent          bool
+	UseDirectComposition bool // Windows DX12 only: set WS_EX_NOREDIRECTIONBITMAP at creation
+	TabbingMode          int
+	TabbingIdentifier    string
+	MinWidth             int // 0 = no minimum constraint
+	MinHeight            int // 0 = no minimum constraint
+	MaxWidth             int // 0 = no maximum constraint
+	MaxHeight            int // 0 = no maximum constraint
+	Icon                 image.Image
 }
 
 // Event represents a platform event.
--- a/internal/platform/platform_windows.go
+++ b/internal/platform/platform_windows.go
@@ -210,6 +210,7 @@ const (
 	// GetSystemMetrics / MonitorFromWindow constants
 	smCXSizeFrame           = 32 // SM_CXSIZEFRAME
 	smCYSizeFrame           = 33 // SM_CYSIZEFRAME
+	smCYCaption             = 4  // SM_CYCAPTION
 	smCXPaddedBorder        = 92 // SM_CXPADDEDBORDERWIDTH
 	monitorDefaultToNearest = 2  // MONITOR_DEFAULTTONEAREST
 
@@ -1041,6 +1042,19 @@ func (w *win32Window) RequestSize(width, height int) {
 
 	outerW := uintptr(outerRect.right - outerRect.left)
 	outerH := uintptr(outerRect.bottom - outerRect.top)
+	if w.frameless {
+		// JBR: WM_NCCALCSIZE removes the top NC area (title bar + top frame
+		// + padded border) and folds it into the client area, but
+		// AdjustWindowRect still included it in the outer height. Subtract it
+		// here so the client area matches the requested logical size.
+		cyCaption, _, _ := procGetSystemMetrics.Call(smCYCaption)
+		cyFrame, _, _ := procGetSystemMetrics.Call(smCYSizeFrame)
+		cyPadded, _, _ := procGetSystemMetrics.Call(smCXPaddedBorder)
+		topNC := cyCaption + cyFrame + cyPadded
+		if topNC < outerH {
+			outerH -= topNC
+		}
+	}
 
 	procSetWindowPos.Call(uintptr(w.hwnd), 0, 0, 0,
 		outerW, outerH,
--- a/internal/platform/platform_windows.go
+++ b/internal/platform/platform_windows.go
@@ -290,9 +290,9 @@ var (
 	procGetMonitorInfoW    = user32.NewProc("GetMonitorInfoW")
 
 	// DWM (Desktop Window Manager) for frameless window shadow
-	dwmapi                        = windows.NewLazyDLL("dwmapi.dll")
-	procDwmExtendFrameIntoClient  = dwmapi.NewProc("DwmExtendFrameIntoClientArea")
-	procDwmFlush                  = dwmapi.NewProc("DwmFlush")
+	dwmapi                       = windows.NewLazyDLL("dwmapi.dll")
+	procDwmExtendFrameIntoClient = dwmapi.NewProc("DwmExtendFrameIntoClientArea")
+	procDwmFlush                 = dwmapi.NewProc("DwmFlush")
 
 	// WaitEvents / WakeUp
 	procMsgWaitForMultipleObjectsEx = user32.NewProc("MsgWaitForMultipleObjectsEx")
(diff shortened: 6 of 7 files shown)


kolkov (changes requested):
Great analysis and thorough A/B testing on DX12+NVIDIA. We verified locally and found a backend compatibility issue.

### What we tested

Checked out the branch on Intel Iris Xe (Win10). Build, tests, lint — all pass (Windows, Linux, macOS). Non-transparent windows render correctly on all 4 backends.

For transparent windows, we wrote a minimal test (`WithTransparent(true) + WithFrameless(true)`) and tested each backend:

| Backend | Result | Why |
|---------|--------|-----|
| **DX12** | ✅ Per-pixel alpha works — desktop visible through semi-transparent pixels | DComp swapchain + NOREDIRECTIONBITMAP = correct path |
| **Vulkan** | ❌ Window fully invisible — no content composited | No DComp; Vulkan uses HWND present. Without DWM redirection surface, present goes to nothing. `resolveAlphaMode` falls back to Opaque (Intel doesn't report `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR` on Win32 surface) |
| **GLES** | ❌ Same as Vulkan | EGL uses HWND, no DComp |
| **Software** | ❌ Same | GDI BitBlt to window without backing surface |

### Root cause

`WS_EX_NOREDIRECTIONBITMAP` disables the DWM redirection surface. This is correct for DX12 which uses DirectComposition (`CreateSwapChainForComposition` + DComp visual tree). But Vulkan/GLES/Software present through the HWND — they need the DWM redirection surface to be visible.

The flag is set at **window creation time** (before backend selection), but only DX12 has the DComp infrastructure to work without a redirection surface.

### How winit handles this

winit exposes `no_redirection_bitmap` as a **separate opt-in** (`with_no_redirection_bitmap(true)`, default false), independent of `transparent`. The wgpu DX12 backend requests it when configuring the surface for DComp. Vulkan/GLES backends don't set it.

For non-`no_redirection_bitmap` transparent windows, winit uses `DwmEnableBlurBehindWindow` with an empty region — the legacy path that works with HWND-based present.

### Proposed solution

Two options (happy to discuss):

**Option A: Defer NOREDIRECTIONBITMAP to surface configuration**
- Window created without NOREDIRECTIONBITMAP (compatible with all backends)
- When DX12 DComp surface is configured, set `WS_EX_NOREDIRECTIONBITMAP` via `SetWindowLongPtrW(GWL_EXSTYLE)` and remove blur-behind
- Other backends: keep `DwmEnableBlurBehindWindow` (legacy transparency, may look different but works)

**Option B: Backend hint at window creation**
- Add a `NeedsNoRedirectionBitmap` flag to surface target or config
- gogpu checks the active backend before setting the style
- Problem: backend isn't selected at window creation time

Option A is simpler — `SetWindowLongPtrW` can change extended styles after creation. @shaolei, what do you think?

### Request: test all backends

Could you verify transparent windows with each backend on your RTX 4060?

```bash
GOGPU_GRAPHICS_API=vulkan   go run your_deskcalendar_test
GOGPU_GRAPHICS_API=dx12     go run your_deskcalendar_test
GOGPU_GRAPHICS_API=gles     go run your_deskcalendar_test
GOGPU_GRAPHICS_API=software go run your_deskcalendar_test
```

On our Intel Iris Xe (Win10), only DX12 shows per-pixel alpha; the other three produce an invisible window. If you see the same on NVIDIA, it confirms the backend-specific issue.

### Minor

```go
slog.Info(...) → slog.Debug(...)
```
Transparent window creation is a routine event, should be Debug level.

kolkov (dismissed):
Great analysis and thorough A/B testing on DX12+NVIDIA. We verified locally and found a backend compatibility issue.

### What we tested

Checked out the branch on Intel Iris Xe (Win10). Build, tests, lint — all pass (Windows, Linux, macOS). Non-transparent windows render correctly on all 4 backends.

For transparent windows, we wrote a minimal test (`WithTransparent(true) + WithFrameless(true)`) and tested each backend:

| Backend | Result | Why |
|---------|--------|-----|
| **DX12** | ✅ Per-pixel alpha works — desktop visible through semi-transparent pixels | DComp swapchain + NOREDIRECTIONBITMAP = correct path |
| **Vulkan** | ❌ Window fully invisible — no content composited | No DComp; Vulkan uses HWND present. Without DWM redirection surface, present goes to nothing. `resolveAlphaMode` falls back to Opaque (Intel doesn't report `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR` on Win32 surface) |
| **GLES** | ❌ Same as Vulkan | EGL uses HWND, no DComp |
| **Software** | ❌ Same | GDI BitBlt to window without backing surface |

### Root cause

`WS_EX_NOREDIRECTIONBITMAP` disables the DWM redirection surface. This is correct for DX12 which uses DirectComposition (`CreateSwapChainForComposition` + DComp visual tree). But Vulkan/GLES/Software present through the HWND — they need the DWM redirection surface to be visible.

The flag is set at **window creation time** (before backend selection), but only DX12 has the DComp infrastructure to work without a redirection surface.

### How winit handles this

winit exposes `no_redirection_bitmap` as a **separate opt-in** (`with_no_redirection_bitmap(true)`, default false), independent of `transparent`. The wgpu DX12 backend requests it when configuring the surface for DComp. Vulkan/GLES backends don't set it.

For non-`no_redirection_bitmap` transparent windows, winit uses `DwmEnableBlurBehindWindow` with an empty region — the legacy path that works with HWND-based present.

### Proposed solution

Two options (happy to discuss):

**Option A: Defer NOREDIRECTIONBITMAP to surface configuration**
- Window created without NOREDIRECTIONBITMAP (compatible with all backends)
- When DX12 DComp surface is configured, set `WS_EX_NOREDIRECTIONBITMAP` via `SetWindowLongPtrW(GWL_EXSTYLE)` and remove blur-behind
- Other backends: keep `DwmEnableBlurBehindWindow` (legacy transparency, may look different but works)

**Option B: Backend hint at window creation**
- Add a `NeedsNoRedirectionBitmap` flag to surface target or config
- gogpu checks the active backend before setting the style
- Problem: backend isn't selected at window creation time

Option A is simpler — `SetWindowLongPtrW` can change extended styles after creation. @shaolei, what do you think?

### Request: test all backends

Could you verify transparent windows with each backend on your RTX 4060?

```bash
GOGPU_GRAPHICS_API=vulkan   go run your_deskcalendar_test
GOGPU_GRAPHICS_API=dx12     go run your_deskcalendar_test
GOGPU_GRAPHICS_API=gles     go run your_deskcalendar_test
GOGPU_GRAPHICS_API=software go run your_deskcalendar_test
```

On our Intel Iris Xe (Win10), only DX12 shows per-pixel alpha; the other three produce an invisible window. If you see the same on NVIDIA, it confirms the backend-specific issue.

### Minor

```go
slog.Info(...) → slog.Debug(...)
```
Transparent window creation is a routine event, should be Debug level.

shaolei:
@kolkov thanks for the detailed review. I dug into Option A and also checked how the rest of the ecosystem actually handles this style. My conclusion is that Option A is not viable: `WS_EX_NOREDIRECTIONBITMAP` is effectively a creation-time style.

**Why `SetWindowLongPtrW` won't save us**

- [microsoft/terminal#11561](https://github.com/microsoft/terminal/issues/11561): "You have to create the window with WS_EX_NOREDIRECTIONBITMAP, SetWindowLong won't work."
- [Stack Overflow #40576394](https://stackoverflow.com/questions/40576394/setwindowlong-with-gwl-exstyle-and-ws-ex-noredirectionbitmap): `SetWindowLong(GWL_EXSTYLE, WS_EX_NOREDIRECTIONBITMAP)` fails with `ERROR_INVALID_PARAMETER` (87). Also, `SetWindowLongPtrW` returns 0 both on failure and on success when the old style value was 0, so this is easy to misread.
- [WebView2Feedback#20](https://github.com/MicrosoftEdge/WebView2Feedback/issues/20): "It is not possible to toggle the WS_EX_NOREDIRECTIONBITMAP ... by SetWindowLongPtr, after the window has been created."
- [Stack Overflow #65868208](https://stackoverflow.com/questions/65868208/wpf-custom-mainwindow-using-createwindowex) (WPF): `SetWindowLong` does nothing for this style.

**Cross-framework check**

| Framework | Flag? | How it is decided |
|---|---|---|
| winit | Yes, but separate opt-in (`with_no_redirection_bitmap`, default false) | Applied to `style_ex` before `CreateWindowEx` ([window_state.rs](https://github.com/rust-windowing/winit/blob/master/winit-win32/src/window_state.rs)); transparent windows without the opt-in use `DwmEnableBlurBehindWindow` with an empty region ([winit#2121](https://github.com/rust-windowing/winit/pull/2121)) |
| wgpu-rs | Yes, DX12 | Requested through the winit `WindowBuilder` before window creation ([3a9113c](https://github.com/gfx-rs/wgpu-rs/commit/3a9113c9622231d63f77274b4f1f63314d140bdf)) — not via a surface-config API; later disabled again ([b588345](https://github.com/gfx-rs/wgpu-rs/commit/b588345568615ef318a109e13ce6bb0961a4b975)) |
| Ebitengine | Yes | Decides before `createNativeWindow` (NoAPI + not transparent + Win10+ + DComp available); the DX driver only reads the flag to pick DComp vs. plain HWND swap chain |
| Godot | Yes, d3d12 only | Decides before `_create_window`; if backend fallback changes the required flag value, it destroys and recreates the window ([display_server_windows.cpp](https://github.com/godotengine/godot/blob/master/platform/windows/display_server_windows.cpp)) |
| Qt | No | Default transparency uses `WS_EX_LAYERED`; the redirection-surface opt-out exists only as `QT_QPA_DISABLE_REDIRECTION_SURFACE` at creation, not as a per-pixel-alpha path |
| SDL3 | No | No current support; [SDL#15776](https://github.com/libsdl-org/SDL/pull/15776) proposes a creation-time hint, and notes `CreateWindowEx` fails with `ERROR_INVALID_PARAMETER` on Windows 7 |
| Flutter | No | Windows transparency is not a supported use case ([#123508](https://github.com/flutter/flutter/issues/123508)); a recent PR removes a transparency helper ([#188186](https://github.com/flutter/flutter/pull/188186)) |
| Bevy | No | Passes through winit; winit's docs say transparency can be set via the builder but not after window creation ([bevy#3032](https://github.com/bevyengine/bevy/issues/3032), [#10929](https://github.com/bevyengine/bevy/issues/10929)) |
| Gio | No | No upstream transparent-window support |

**Conclusion**

None of the frameworks we checked toggles this flag after `CreateWindowEx`. The ones that use it decide at creation — either from an explicit user opt-in (winit) or from a backend that is already resolved (Godot/Ebitengine); if the resolved backend changes the required value, Godot recreates the window.

So Option A cannot be the fix. Option B's timing problem is real for `GraphicsAPIAuto`, but note gogpu already knows `Config.GraphicsAPI` before `CreateWindow`, so a forced DX12 transparent window could be decided at creation. For Auto, the viable paths seem to be:

1. Resolve the backend before creating the window.
2. Recreate the window if the selected backend differs (Godot's approach).
3. Split `Transparent` from an explicit no-redirection opt-in (winit's model), keeping `DwmEnableBlurBehindWindow` as the legacy default.

One small correction to the review: wgpu requested this style through the winit builder before window creation, not when configuring the surface.

For this PR, the narrowest safe direction seems to be to keep the flag off for non-DX12 (via backend resolution before creation, or by documenting per-pixel alpha as DX12-only); otherwise we regress Vulkan/GLES/Software transparent windows from "legacy transparency" to "invisible window". Do you think window recreation on backend fallback is acceptable, or would you prefer the winit-style explicit opt-in? Is there another approach you would suggest for the Auto case?

kolkov:
@shaolei — outstanding research on `SetWindowLongPtrW`. You're right: Option A is dead. We did further research across the ecosystem and found a clearer picture.

## Key finding: `DwmBlurBehind` enables per-pixel alpha for Vulkan too

The [jeweg/win32-window-transparency](https://github.com/jeweg/win32-window-transparency) repository provides reference demos for per-pixel alpha on Windows 10/11. Their recommended setup for Vulkan transparency (program `07_vulkan_perpixel_alpha`):

1. `BLACK_BRUSH` background brush
2. `DwmEnableBlurBehindWindow` with `DWM_BB_ENABLE | DWM_BB_BLURREGION` + empty region via `CreateRectRgn(0, 0, -1, -1)`
3. Vulkan swapchain with `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR`

**No `WS_EX_NOREDIRECTIONBITMAP`.** The README explicitly warns: *"Removing the redirection bitmap with `WS_EX_NOREDIRECTIONBITMAP` can cause swapchain content to not appear at all."*

This means the `DwmBlurBehind` code your PR removes is actually the **correct universal path** — it enables per-pixel alpha for Vulkan and potentially other HWND-based backends, as long as the GPU driver reports `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR` support.

**Caveat:** not all Vulkan drivers report premultiplied alpha on Win32 surfaces. Our `resolveAlphaMode` falls back to Opaque when it's unsupported — in that case, transparency is not possible on Vulkan regardless of the window setup.

## Enterprise consensus: NOREDIRECTIONBITMAP = DComp only

Every project that uses `WS_EX_NOREDIRECTIONBITMAP` gates it on DirectComposition:

| Project | Approach | Source |
|---------|----------|--------|
| **Mozilla Firefox** | "Only use `WS_EX_NOREDIRECTIONBITMAP` when using DirectComposition" — resolved fixed in Firefox 74 | [Bug 1610949](https://bugzilla.mozilla.org/show_bug.cgi?id=1610949) |
| **VLC** | `WS_EX_NOREDIRECTIONBITMAP` only in `CompositorDirectComposition` | [vlc-commits](https://www.mail-archive.com/vlc-commits@videolan.org/msg68335.html) |
| **Godot** | DX12 only; recreates window on backend fallback | (from your research) |
| **Ebitengine** | Decides before `createNativeWindow`; DComp path only | (from your research) |
| **winit** | Separate opt-in (`with_no_redirection_bitmap`), default=off; transparent without opt-in uses `DwmBlurBehind` | [winit source](https://github.com/rust-windowing/winit/blob/master/winit-win32/src/window.rs#L1282) |
| **jeweg reference** | Explicitly warns against NOREDIRECTIONBITMAP; uses `DwmBlurBehind` for Vulkan | [README](https://github.com/jeweg/win32-window-transparency) |

Related wgpu issues: [#7108](https://github.com/gfx-rs/wgpu/issues/7108) (DX12 DComp transparency), [#1375](https://github.com/gfx-rs/wgpu/issues/1375) (DX12 alpha), [#3486](https://github.com/gfx-rs/wgpu/issues/3486) (portable window transparency), [egui#2680](https://github.com/emilk/egui/issues/2680) (eframe wgpu transparency). SDL3: [#15751](https://github.com/libsdl-org/SDL/issues/15751) (transparency over DXGI), [#15776](https://github.com/libsdl-org/SDL/pull/15776) (creation-time hint proposed).

## Our recommended approach: two-tier transparency

**Tier 1 — `DwmBlurBehind` (universal default):**
- Works for Vulkan (with `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR`), DX12 HWND path, GLES
- The code your PR removes — it should **stay** as the default transparent window path
- Mozilla, jeweg, winit all use this as the baseline

**Tier 2 — `WS_EX_NOREDIRECTIONBITMAP` + DComp (DX12 optimization):**
- Better compositing quality — true DirectComposition, no DWM intermediary
- Requires creation-time decision
- Your PR's NOREDIRECTIONBITMAP + skip-DwmExtendFrame + skip-BlurBehind — correct for this tier only

**Critical constraint for Auto mode:** our backend priority is **Vulkan > DX12 > GLES**. So `Auto + dcompAvailable()` → NOREDIRECTIONBITMAP is **unsafe** — wgpu will pick Vulkan first, not DX12, and the window becomes invisible. This rules out optimistic DComp for Auto.

**Decision matrix:**

| `config.GraphicsAPI` | `config.Transparent` | Window style | Transparency mechanism |
|---------------------|---------------------|-------------|----------------------|
| DX12 (explicit) | true | `WS_EX_NOREDIRECTIONBITMAP` | DComp (your PR) |
| Vulkan/GLES/Software | true | normal | `DwmBlurBehind` (current code) |
| Auto | true | normal | `DwmBlurBehind` — safe for any backend |
| any | false | normal | no transparency |

**Implementation:** pass `GraphicsAPI` hint into `platform.Config`. In `createWindowWin32`:

```go
useDComp := config.Transparent && config.GraphicsAPI == GraphicsAPIDX12

if useDComp {
    dwExStyle = wsExNoRedirectionBitmap
    // skip DwmBlurBehind, skip DwmExtendFrame for transparent
} else if config.Transparent {
    enableDwmBlurBehind(hwnd)  // universal fallback — works for Vulkan+premultiplied
}
```

For `Auto`, we default to `DwmBlurBehind` — safe for whichever backend wgpu selects. Users who specifically want DComp quality can set `WithGraphicsAPI(GraphicsAPIDX12)` + `WithTransparent(true)`.

**What changes from your current PR:**
- ✅ Keep `WS_EX_NOREDIRECTIONBITMAP` logic — but gate on `GraphicsAPI == DX12`
- ✅ Keep skip-`DwmExtendFrameIntoClientArea` for transparent+frameless
- ✅ Keep frameless `SetSize` NC offset fix (commit `e8e3160`)
- ❌ Don't delete `enableDwmBlurBehind` — it's the Vulkan/GLES/Auto path
- ➕ Add `GraphicsAPI` field to `platform.Config`

What do you think? If you have time, testing `DwmBlurBehind` + Vulkan + premultiplied alpha on your RTX 4060 would confirm the jeweg reference (NVIDIA may or may not report `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR` on Win32 surfaces).

shaolei:
@kolkov this two-tier direction looks right and matches the ecosystem. Two caveats before we lock the decision matrix:

**1. Auto + `DwmBlurBehind` is "safe" but not equivalent to per-pixel alpha when Auto picks DX12.**

In the wgpu dependency used here (v0.30.35), the DX12 backend chooses DirectComposition based on the requested alpha mode, not on the window style: [`hal/dx12/surface.go`](https://github.com/gogpu/wgpu/blob/v0.30.35/hal/dx12/surface.go) does `useDComp := config.AlphaMode == CompositeAlphaModePremultiplied` and notes that `CreateSwapChainForHwnd` only supports `DXGI_ALPHA_MODE_IGNORE`. Our `resolveAlphaMode` requests Premultiplied for every transparent window when the adapter advertises it, so an Auto window that resolves to DX12 still gets a DComp swap chain — and without `WS_EX_NOREDIRECTIONBITMAP` that is exactly the original #361 symptom (semi-transparent content composites against the opaque redirection surface instead of the desktop). Auto+blur is therefore safe against "invisible window", but not a working per-pixel path on Vulkan-less machines where Auto falls through to DX12. That corner case should be documented, or handled by recreating the window when Auto selects DX12.

**2. Software backend likely needs more than `DwmBlurBehind`.**

The software HAL presents through GDI `BitBlt`/`StretchDIBits` (`SRCCOPY`) into the window DC ([`blit_windows.go`](https://github.com/gogpu/wgpu/blob/v0.30.35/hal/software/blit_windows.go)); GDI does not blend per-pixel alpha, so `DwmBlurBehind` alone may still produce opaque content. This needs a real test before we claim Software works; if it fails, the options are `AlphaBlend` from a DIB section, `UpdateLayeredWindow`, or documenting transparent+Software as unsupported.

With those two caveats, the matrix I would implement:

| API | transparent | window style | expected |
|---|---|---|---|
| DX12 (explicit) | yes | `WS_EX_NOREDIRECTIONBITMAP` | per-pixel via DComp (this PR) |
| Vulkan (explicit) | yes | normal + blur | per-pixel if `PRE_MULTIPLIED` is reported |
| GLES (explicit) | yes | normal + blur | likely per-pixel (EGL) |
| Software (explicit) | yes | normal + blur | needs testing (GDI) |
| Auto | yes | normal + blur | Vulkan/GLES per-pixel; DX12 falls back to #361 behavior |

On implementation, I would keep the platform layer backend-agnostic and pass a `UseDirectComposition bool` (computed in `app.go` as `transparent && GraphicsAPI == DX12`) into `platform.Config` instead of the full `GraphicsAPI` type; either works, but the bool avoids coupling `internal/platform` to `gpu/types`. Everything else in your plan — keep `enableDwmBlurBehind`, gate NOREDIRECTIONBITMAP on explicit DX12, skip `DwmExtendFrameIntoClientArea` for transparent+frameless — matches what I would do.

I can run the Vulkan/GLES/Software matrix on the RTX 4060 to confirm the per-backend behavior before implementing.

kolkov:
@shaolei — we did deep research on both your caveats. The findings simplify the picture significantly.

## Caveat 1: DwmBlurBehind + DComp actually works WITHOUT WS_EX_NOREDIRECTIONBITMAP

`DwmEnableBlurBehindWindow` with an empty region makes the DWM redirection surface transparent. DComp visual then composites **through** the transparent surface against the desktop. This is exactly the winit pattern ([window.rs:1282](https://github.com/rust-windowing/winit/blob/master/winit-win32/src/window.rs#L1282)) — if `no_redirection_bitmap` is false, winit calls `DwmBlurBehind` and DComp still works.

This raises the question: **why did #361 fail if our code already called `enableDwmBlurBehind`?**

Our hypothesis: the root cause of #361 was not the absence of `WS_EX_NOREDIRECTIONBITMAP`, but **`DwmExtendFrameIntoClientArea` overriding `DwmBlurBehind`** for frameless windows. Your own commit `8c41d1c` says: *"it switches the window onto the DWM glass composition path, which overrides DirectComposition per-pixel alpha."*

In the current code, for frameless+transparent windows, BOTH are called in `createWindowWin32`:
- Line 819: `DwmExtendFrameIntoClientArea` (frameless shadow)
- Line 829: `enableDwmBlurBehind` (transparency)

Even though BlurBehind is called second, `DwmExtendFrame` may put the window into a DWM glass composition state that BlurBehind cannot undo. This matches your observation perfectly.

**Could you test this specific scenario?** On your Win11 + RTX 4060:

1. Revert `WS_EX_NOREDIRECTIONBITMAP` (use `dwExStyle = 0`)
2. Keep `DwmBlurBehind` (don't delete `enableDwmBlurBehind`)
3. Keep skip-`DwmExtendFrame` for transparent+frameless (your commit `8c41d1c`)
4. Test DeskCalendar with DX12

If transparency works → the entire NOREDIRECTIONBITMAP complexity is unnecessary. The fix is simply: skip `DwmExtendFrame` for transparent windows + keep `DwmBlurBehind`.

## Caveat 1b: Vulkan transparency via DwmBlurBehind + OPAQUE swapchain

Major finding: **DwmBlurBehind enables per-pixel alpha for Vulkan even with `VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR`.**

DWM reads the alpha channel from the redirection surface regardless of the Vulkan swapchain's `compositeAlpha` setting. This means our `resolveAlphaMode` falling back to OPAQUE is fine — DwmBlurBehind handles transparency at the DWM level.

Driver coverage for `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR`:

| Vendor | PRE_MULTIPLIED on Win32 | Source |
|--------|:---:|--------|
| **NVIDIA** (~60% market) | ❌ reports only OPAQUE | [NVIDIA Forum](https://forums.developer.nvidia.com/t/no-support-for-vk-composite-alpha-inherit-bit-khr/77417) |
| **AMD** (~15%) | ❌ broken since 23.10 | [wgpu#5368](https://github.com/gfx-rs/wgpu/issues/5368) |
| **Intel** (~25%) | ✅ only vendor | [Khronos Forum](https://community.khronos.org/t/transparent-window-possible-with-integrated-graphics-only/104237) |

So PRE_MULTIPLIED covers ~0-15% of users. But DwmBlurBehind + OPAQUE covers **100%**. The [jeweg/win32-window-transparency](https://github.com/jeweg/win32-window-transparency) reference repo confirms this pattern works.

## Caveat 2: Software backend — confirmed, defer

GDI `BitBlt(SRCCOPY)` does not preserve alpha — GDI clears it to 0x00. `DwmBlurBehind` alone won't help. The fix is `UpdateLayeredWindow` with `WS_EX_LAYERED` (Qt6 `QPlatformBackingStore::flush()` pattern, ~25 LOC). SDL3 does not support transparent+software. Reasonable to defer and document as a known limitation.

## Revised recommendation

If the DwmBlurBehind + skip-DwmExtendFrame test passes on your hardware, the PR simplifies dramatically:

| Change | Keep/Remove |
|--------|-------------|
| Skip `DwmExtendFrameIntoClientArea` for transparent+frameless | ✅ Keep — this was the actual #361 root cause |
| `enableDwmBlurBehind` | ✅ Keep — universal transparency for all backends |
| `WS_EX_NOREDIRECTIONBITMAP` | ❌ Remove — unnecessary if DwmBlurBehind + skip-DwmExtendFrame works |
| Delete blur-behind helpers | ❌ Don't delete — they're the universal path |
| Frameless `SetSize` NC offset fix | ✅ Keep — unrelated, good fix |

No `GraphicsAPI` in `platform.Config`. No two-tier system. No backend coordination. Just: skip `DwmExtendFrame` for transparent + keep `DwmBlurBehind`.

**But this needs your empirical verification.** Our analysis is from code + documentation + enterprise references. Your A/B testing on real hardware is what confirmed #361 in the first place — we need the same rigor to confirm this simpler path works.

shaolei:
@kolkov — we ran the exact A/B you asked for on Windows 11 + RTX 4060 (driver 591.86). The simplified path does not work for DX12.

**Test setup**

Frameless 360×480 transparent window, full-screen black backdrop, two phases: opaque red `(1,0,0,1)` and 50% red `(0.5,0,0,0.5)`. We captured the window center from the screen and also observed the corners with a magenta boundary marker.

| Configuration | Opaque phase | 50% alpha phase | Interpretation |
|---|---|---|---|
| `dwExStyle=0` + `DwmBlurBehind` + skip `DwmExtendFrame` (your revised proposal) | 255,0,0 | **255,128,128** | 50% red blended against an **opaque white redirection surface**; desktop is not visible |
| `WS_EX_NOREDIRECTIONBITMAP` + DComp, no blur-behind (two-tier) | 255,0,0 | ~247,120,120 over gray backdrop | ~50% blend with the background; the four corners are transparent and desktop shows through |

During the no-flag run, resizing the left edge left produced a white strip on the right, and resizing the right edge right produced a transparent strip — the DComp-swapchain-vs-redirection-surface mismatch that #361 was about. Skipping `DwmExtendFrame` does not fix it.

**Your caveat 1b is confirmed**

Vulkan + `DwmBlurBehind` + forced `VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR` works: opaque 255,0,0 → 50% 127,0,0 over black. DWM-level blur handles alpha even with an OPAQUE swapchain.

One correction to the vendor table: on this machine NVIDIA **does** report `VK_COMPOSITE_ALPHA_PRE_MULTIPLIED_BIT_KHR` (driver 591.86), and that path also worked. The "NVIDIA reports only OPAQUE" entry is not true for this driver.

**Software**

Confirmed broken: GDI `BitBlt`/`StretchDIBits` (`SRCCOPY`) does not preserve per-pixel alpha. We've documented it as unsupported and deferred the `UpdateLayeredWindow` follow-up.

**Current implementation**

We kept the two-tier model:

- Explicit DX12 + transparent → `WS_EX_NOREDIRECTIONBITMAP` at creation, no `DwmBlurBehind`.
- Vulkan/GLES/Auto/Software + transparent → `DwmBlurBehind`.
- Transparent + frameless skips `DwmExtendFrameIntoClientArea`.
- Frameless `RequestSize` NC-offset fix kept.
- Instead of adding `GraphicsAPI` to `platform.Config`, we pass a `UseDirectComposition bool` computed in the app layer (`transparent && GraphicsAPI == DX12`), which avoids coupling `internal/platform` to `gpu/types`.

For Auto + DX12 we're documenting the limitation rather than recreating the window for now.

Do you agree with keeping the two-tier approach, or do you have another suggestion for the Auto case?

kolkov (approved):
Excellent work. Two-tier transparency architecture validated by A/B testing on real hardware. Merging — minor follow-ups (HiDPI GetSystemMetricsForDpi, import alias) tracked internally.

Issue #418, opened 2026-08-02
docs(examples): add a simple fractal renderer as an example GoGPU application


jbunds:
I wrote a simple Mandelbrot set renderer using gogpu which automatically zooms in on preset x, y coordinates.

I would like to submit it to `examples/mandelbrot`, but I need some help from a domain expert to review the code.

Please advise regarding how I should proceed.

jbunds:
I published it at https://github.com/jbunds/mandelbrot/tree/main/gogpu and verified it works using Go 1.25.0 on a MacBook Air M1.

It's MIT-licensed, so anyone can freely copy and modify it.

kolkov:
@jbunds — great work! This is a solid GPU compute example — emulated double-precision in WGSL, palette precompute, direct `textureStore` into the surface. Exactly the kind of showcase we want in our examples.

### Where it belongs

`gogpu/examples/mandelbrot/` — confirmed. It demonstrates direct GPU compute via `wgpu.Device` + `ComputePipeline`, which is core gogpu territory.

### How to proceed

Please open a PR adding your code to `examples/mandelbrot/`. A few notes for the PR:

1. **go.mod** — up to you, but for reference none of the existing examples in this repo have their own `go.mod` (they build as part of the main module). Your imports are all stdlib + gogpu ecosystem, so it should work either way.

2. **Resource cleanup** — consider releasing `paletteBuf`, `uniformBuf`, `bgLayout0`, `bgLayout1`, `pipeline` in `OnClose` alongside `staticBindGroup`. Not critical for an example, but good practice.

3. **The `stopRendering` flag** — your TODO about `StartAnimation` is on the right track. With our event-driven model, you could use `token := app.StartAnimation()` at start and `token.Stop()` at frame 2745 instead of `WithContinuousRender(true)` + bool flag.

None of these are blockers — happy to iterate on the PR.

Looking forward to it!

kolkov:
Merged in #438 (Mandelbrot fractal renderer example).

PR #438, opened 2026-08-08
docs(examples): add a Mandelbrot set fractal renderer demo (#418)

jbunds:
Please take a look.


jbunds:
Should I commit `go.mod` and `go.sum` to `examples/mandelbrot` on the dev branch to fix the compilation errors?

kolkov:
No separate `go.mod` needed — our examples are part of the main module (like `examples/triangle/`, `examples/drag_source/`, etc.). The compilation errors are likely from formatting or missing imports. Check CI logs for the exact errors:

```bash
gh run view --log-failed
```

Also make sure to run locally before push:
```bash
gofmt -w examples/mandelbrot/
go build ./examples/mandelbrot/
golangci-lint run --timeout=5m ./examples/mandelbrot/
```

porjo:
@jbunds you should run `go mod tidy` in the repo root to add the additional packages your example needs, then include the changes to `go.mod` and `go.sum` with your PR.

jbunds:
> @jbunds you should run `go mod tidy` in the repo root to add the additional packages your example needs, then include the changes to `go.mod` and `go.sum` with your PR.

I personally prefer not to pollute the root module with the following packages that only my example program requires:

```
  github.com/gogpu/gg
  github.com/gogpu/gg/gpu
  github.com/gogpu/gg/integration/ggcanvas
```
  
But if the repo owners feel strongly that i should change the root `go.mod` and `go.sum` files to include those packages, then I will do so.

In the latest commit to the dev branch I added `go.mod` and `go.sum` to `examples/mandelbrot`, and that rectified the compiler and linter errors while maintaining a minimal set of dependencies for the `github.com/gogpu/gogpu` package.

kolkov:
Great work on the Mandelbrot demo, @jbunds! The double-precision emulation in WGSL (Dekker's algorithm) and the native menu system for switching points of interest are impressive. Code quality is high overall.

A few items to fix before merge:

### Must fix

1. **go.mod module path** — currently `github.com/gogpu/examples/mandelbrot`, should be `github.com/gogpu/gogpu/examples/mandelbrot` (the repo is `gogpu/gogpu`, not `gogpu/examples`).

2. **BufferBindingType** — `resources.go:85`: the palette buffer is declared as `var<storage, read>` in WGSL, so the Go-side layout should use `BufferBindingTypeReadOnlyStorage` instead of `BufferBindingTypeStorage`. Matches our `particles` example pattern and enables better driver optimizations.

3. **Import ordering** — `resources.go`: stdlib imports (`math`) must come before third-party imports. Move `"math"` above the `gogpu/` imports with a blank line separator.

### Recommended (non-blocking)

4. **Dead WGSL code** — `primitive_sum` function in `mandelbrot.wgsl` is defined but never called. Remove it.

5. **FPS integer division** — `main.go:135`: `float64(1000 / elapsed)` does integer division first. Use `1000.0 / float64(elapsed)` for accurate FPS.

6. **Nil guard** — `main.go:239`: `token.Load().Stop()` can panic if animation already stopped. Add `if t := token.Load(); t != nil { t.Stop() }`.

7. **Unnecessary goroutine** — `RequestRedraw()` can be called directly from OnDraw, no need for `go func() { runtime.Gosched(); ... }()`.

8. **Magic number 2745** — consider making it a named constant with a comment explaining the double-double precision limit.

Build passes, lint is clean on all platforms. Looking forward to merging this after the fixes!

kolkov (approved):

PR #440, opened 2026-08-09
feat(win32): add KeyCancel for Ctrl+Break (VK_CANCEL)

Zeroes1:
On Win32 the Pause key (VK_PAUSE) and Ctrl+Break (VK_CANCEL) are two
distinct virtual keys, but vkCodeToKey reported both as KeyPause, so
applications could not tell an interrupt from the plain Pause key.

Add a dedicated KeyCancel so the distinction survives to the public
input API. Depends on gpucontext PR adding KeyCancel to the Key enum.

- internal/platform/platform_windows.go: VK_CANCEL -> gpucontext.KeyCancel
- input/keyboard.go: KeyCancel in the public Key enum
- app.go: gpucontext.KeyCancel -> input.KeyCancel in mapToInputKey
- app_test.go: mapping case ({"Cancel", ...})

Handling Ctrl+Break (Ctrl+Pause) as an interrupt is useful because:

It is the second canonical stop combination on Windows. Alongside Ctrl+C, Ctrl+Break is the standard way to interrupt execution in a console (on Unix: intr/quit, SIGINT/SIGQUIT). The GUI build of f4 used to lose this pair — gogpu either reported it as Pause or dropped it entirely.
Not everything stops with Ctrl+C. Many interactive programs (vim, ssh, various CLI tools) trap Ctrl+C or misbehave on it. Ctrl+Break is the fallback/alternative interrupt users expect from a terminal out of the box.
A GUI window has no console signal. In a real console, the OS handles Ctrl+Break. Inside a gogpu window there is no console window, so the combination must reach the PTY as ordinary input — through the same path as Ctrl+C (ETX) — transplanting familiar terminal behavior into the GUI.
It unblocks hung commands. Without it, a long-running operation (dir /s, a full-disk grep, an interactive shell) in the fullscreen GUI build could only be stopped by closing the window.
The cost is minimal and safe. Via a dedicated KeyCancel code, the Pause key keeps its own meaning (media/scroll), and only "Pause + Ctrl" becomes the interrupt — preserving both compatibility and API distinguishability.

### DIFF
--- a/app.go
+++ b/app.go
@@ -2014,6 +2014,8 @@ func gpucontextKeyToInputKey(key gpucontext.Key) input.Key {
 		return input.KeyNumLock
 	case gpucontext.KeyPause:
 		return input.KeyPause
+	case gpucontext.KeyCancel:
+		return input.KeyCancel
 
 	default:
 		return input.KeyUnknown
--- a/app_test.go
+++ b/app_test.go
@@ -363,6 +363,7 @@ func TestGpucontextKeyToInputKey(t *testing.T) {
 		{"ScrollLock", gpucontext.KeyScrollLock, input.KeyScrollLock},
 		{"NumLock", gpucontext.KeyNumLock, input.KeyNumLock},
 		{"Pause", gpucontext.KeyPause, input.KeyPause},
+		{"Cancel", gpucontext.KeyCancel, input.KeyCancel},
 		{"Unknown", gpucontext.Key(9999), input.KeyUnknown},
 	}
 
--- a/input/keyboard.go
+++ b/input/keyboard.go
@@ -131,6 +131,7 @@ const (
 	KeyPrintScreen
 	KeyScrollLock
 	KeyPause
+	KeyCancel // Win32 Ctrl+Break (VK_CANCEL); distinct from KeyPause
 
 	KeyCount // Number of keys
 )
--- a/internal/platform/platform_windows.go
+++ b/internal/platform/platform_windows.go
@@ -2003,6 +2003,7 @@ const (
 	vkControl   = 0x11
 	vkMenu      = 0x12 // Alt
 	vkPause     = 0x13
+	vkCancel    = 0x03 // Ctrl+Break arrives as VK_CANCEL, not VK_PAUSE
 	vkCapital   = 0x14 // Caps Lock
 	vkSpace     = 0x20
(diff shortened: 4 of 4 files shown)


kolkov:
Thanks for the clean PR, @Zeroes1! The KeyCancel mapping and test look correct.

Good news: we just released **gpucontext v0.26.0** which already includes `KeyCancel` (value 266) as part of a Key enum redesign (grouped explicit bases, 138 keys total). So no companion gpucontext PR is needed — it's already there.

To get this PR mergeable, please:

1. **Rebase onto latest `main`** — we'll bump gogpu's `go.mod` to `gpucontext v0.26.0` shortly as part of our cascade release. After that, rebase and the build will find `gpucontext.KeyCancel`.

2. **Test locally** — verify the VK_CANCEL → KeyCancel path works in your f4 terminal (Ctrl+Break sends the expected key event).

3. **(Optional) X11 mapping** — we noticed `KeysymCancel = 0xff69` is already defined in `internal/platform/x11/keyboard.go:49` but never mapped to `gpucontext.KeyCancel`. If you'd like to add the one-liner X11 case in the same PR, that would be great. Otherwise we'll track it as a follow-up.

4. **Import alias** — minor: the test uses `gpu_types` (snake_case). Go convention is `gputypes` or just `types`.

We'll ping you once the dep bump lands on main. Thanks for contributing — looking forward to seeing f4!

Zeroes1:
Thanks for the review, kolkov!

I didn't wait for the go-ahead and went ahead with the rebase — the force-push updated the branch onto the latest main (v0.51.0, gpucontext v0.26.0). Local build and tests pass, and the CI should now pick up gpucontext.KeyCancel.

Regarding the X11 mapping: I skipped it. Ctrl+Break isn't really a used combo in the Linux ecosystem — terminals there rely on Ctrl+C / interrupt semantics instead, so mapping KeysymCancel would add a code path with no practical use case. I'd rather leave it as a separate follow-up if someone actually needs it.

On the import alias: the gpu_types import in app_test.go was already there before my PR, so I didn't touch it to keep the diff focused on the KeyCancel change. Happy to rename it to gputypes if you'd still like that in a separate change.

Zeroes1:
locally tested with code f4, it works!

kolkov (approved):

Issue #442, opened 2026-08-10
bug: Debug overlay system (ADR-066) — bind group crash + frozen animation


kolkov:
### Environment
- gogpu v0.51.0, gg v0.51.0, gpucontext v0.26.0, wgpu v0.31.0
- ui v0.1.51 (feat/damage-overlay-adr065-066 branch)
- Windows 10, Intel Iris Xe, Vulkan backend

### Bug 1: `GOGPU_DEBUG_DIRTY=overlay` — bind group crash every frame

**Symptom:** Every frame produces:
```
ERROR submit failed err="wgpu: Submit: command buffer at index 0 references released bind group"
```

**Root cause:** The dirty widget overlay (`ui/desktop/debug_dirty.go:55-83`) stores a reference to the SAME `gg.Context` used by the main UI render path. When the overlay's `Draw()` calls `FlushGPUWithViewPreserveContent`, gg's `GPURenderContext.Flush()` sees `sharedEncoder == nil` (cleared after main render returned via `defer cc.SetSharedEncoder(gpucontext.CommandEncoder{})` in `gg/integration/ggcanvas/canvas.go:641`).

This causes gg to:
1. Create its own command encoder (independent submit path, `render_session.go:3221-3296`)
2. Submit that encoder immediately
3. Run deferred `releasePendingBindGroups()` (`render_session.go:942`) which releases bind groups accumulated during the MAIN render pass

These bind groups (`gpuTexBindGroups`, `surfaceCompositeBindGroup`) are still referenced by gogpu's shared frame encoder, which submits AFTER `drawDebugOverlays` returns (`renderer.go:610`).

**Why damage overlay works:** Uses `overlayPipeline.renderInstances()` on the SHARED encoder with its own bind groups. No gg involvement.

**Recommended fix:** Migrate dirty overlay to built-in `overlayPipeline` (same as damage overlay). Both draw flat-color rectangles — the instanced quad pipeline handles this. Eliminates gg dependency and bind group cross-contamination.

Alternatively: give the dirty overlay its own dedicated `gg.Context` (not the shared UI context).

### Bug 2: `GOGPU_DEBUG_DAMAGE=overlay` — fade animation freezes on idle

**Symptom:** Green damage overlay renders on first frame, but fade animation stops until next user interaction. Frames don't tick during idle.

**Root cause:** In `app.go:1129-1136`:
```go
if !ws.frameStarted {
    if ws.acquireFailed {
        a.RequestRedraw()
    }
[...]

PR #444, opened 2026-08-10
fix: continue debug overlay animations on idle frames

besmpl:
## Summary

- keep debug-overlay animation demand pending across lazy-frame cleanup
- acquire an overlay-only frame when application content is idle
- propagate clear, submit, and present completion so failed overlay-only frames park work instead of self-scheduling forever
- preserve the existing content-frame retry policy and defer overlay work on pixel-present paths
- clear stale demand when the last overlay is removed
- add lifecycle coverage for idle animation, acquire/submit/present failures, unavailable surfaces, pixels, and multiple surfaces

## Why

An overlay can request its next animation frame after the application has no content damage. The lazy frame path previously discarded that demand before acquiring or presenting another frame, so animations stopped after the first draw. Merely preserving the flag is not sufficient: persistent surface or queue failures would then create an unbounded redraw loop. This change carries completion status through the frame boundary and only self-schedules a successful overlay-only frame.

## Verification

- focused overlay lifecycle tests, repeated and under the race detector
- debug overlay tests with damage/FPS environment modes enabled
- `go test ./...`
- `go test -race ./...`
- Linux and Windows package test cross-compilation
- `gofmt` and `git diff --check`

`go vet ./...` reports only existing Darwin `unsafe.Pointer` warnings in unchanged platform files.

Addresses Bug 2 of #442. Bug 1 remains outside this PR.

### DIFF
--- a/app.go
+++ b/app.go
@@ -1113,6 +1113,15 @@ func (a *App) renderFrameGPU(frames []windowFrame) {
 		ctx := newContextForSurface(a.renderer, ws, frame.scale)
 		frame.onDraw(ctx)
 
+		// An overlay that requested another frame is pending GPU work even when
+		// application content is clean. Start an overlay-only frame so the normal
+		// end-frame path can draw, submit, and present it.
+		overlayOnlyAcquire := !ws.frameStarted && !ws.pixelPresented &&
+			!ws.acquireFailed && ws.overlayNeedsRedraw
+		if overlayOnlyAcquire {
+			ws.ensureFrameStarted()
+		}
+
 		// No frame started. Two very different reasons:
 		//
 		// The callback issued no draw calls, so lazy acquire never acquired
@@ -1127,7 +1136,11 @@ func (a *App) renderFrameGPU(frames []windowFrame) {
 		// (surface outdated, not yet configured). Demand-driven mode already
 		// consumed the invalidation, so that one does need another frame.
 		if !ws.frameStarted {
-			if ws.acquireFailed {
+			// Content acquisition keeps the existing retry policy because its
+			// invalidation was consumed. Overlay-only work remains pending until
+			// the next external invalidation instead of spinning while a surface
+			// is minimized, unconfigured, or lost.
+			if ws.acquireFailed && !overlayOnlyAcquire {
 				a.RequestRedraw()
 			}
 			ws.resetLazyState()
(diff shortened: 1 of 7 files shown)


kolkov:
@besmpl — thank you for the thorough analysis and clean implementation. The overlay lifecycle issues you identified are exactly right.

However, we've since taken a different architectural direction that supersedes this approach. After deep enterprise research (Chromium cc/, GTK4 GSK, Flutter flow/), we discovered that the overlay freeze was a symptom of a fundamental ownership inversion: the drawing library (gg) owned compositor resources (composition textures, LoadOp decisions, blit pipeline) that belong at the compositor level (gogpu).

**ADR-067** ([gogpu#443](https://github.com/gogpu/gogpu/issues/443), [gg#493](https://github.com/gogpu/gg/issues/493)) restructures the rendering architecture:
- gogpu owns a composition texture — content renderers draw into it
- gogpu's compositor decides Clear/Load, not gg
- `SurfaceCompositor` interface in gpucontext bridges the boundary
- Overlay-only frames work because the compositor preserves composition texture content
- `selectBaseLayer` uses `target.PreserveContent` (explicit external content signal), not `frameRendered` (which was set by any Flush)

This is a larger change touching gogpu, gg, and gpucontext — currently in local commits, not yet released. The overlay lifecycle fix you proposed would conflict with these changes since the frame loop and `overlayNeedsRedraw` handling are significantly restructured.

We'll close this PR once ADR-067 is released. Your analysis of the idle-frame problem was instrumental in understanding the deeper architectural issue — the overlay freeze led us to discover the compositor ownership inversion.

Once the refactoring lands, we'd appreciate your review and validation — especially the overlay lifecycle, frame scheduling, and multi-renderer composition paths. Your expertise on the frame loop and overlay timing would be invaluable for catching edge cases we might have missed.

kolkov:
**ADR-067 is now released across the ecosystem.** Cascade:

| Repo | Version | What changed |
|------|---------|-------------|
| [gpucontext v0.27.0](https://github.com/gogpu/gpucontext/releases/tag/v0.27.0) | `SurfaceCompositor` interface |
| [gogpu v0.52.0](https://github.com/gogpu/gogpu/releases/tag/v0.52.0) | Compositor-owned render target, `composView`, blit + composite pipelines |
| [gg v0.52.0](https://github.com/gogpu/gg/releases/tag/v0.52.0) | Pure drawing library — compositor decisions delegated to gogpu |
| [ui v0.1.52](https://github.com/gogpu/ui/releases/tag/v0.1.52) | Pluggable DebugOverlay (ADR-066) + deps cascade |
| [g3d v0.1.7](https://github.com/gogpu/g3d/releases/tag/v0.1.7) | Damage source registration (ADR-065) + deps cascade |

The overlay freeze you diagnosed is fixed by the compositor architecture — overlay-only frames now work because gogpu preserves the composition texture and blits it to swapchain, then draws overlay on top. No more alpha accumulation, no more freeze on idle content.

@besmpl — would appreciate your validation, especially the overlay lifecycle and frame scheduling paths. Your analysis was the trigger for this whole redesign. Since the architecture has changed significantly, please review your other open PRs across the ecosystem and close any that now conflict with the new compositor ownership model.

@lkmavi — if you get a chance, please validate on macOS (Metal). Key scenarios: gg-only rendering, overlay fade on idle content, transparent window.

besmpl:
Thanks for the update. Closing as superseded by the released ADR-067 compositor-owned render target architecture in v0.52.0; this branch targets the previous frame-loop ownership model and should not be rebased into the new design.

PR #452, opened 2026-08-12
feat(wayland): add one-shot presentation sync

kivutar:
## Summary

- add Context.RequestPresentationSync for infrequent compositor-paced frames
- let Wayland force one wl_surface.frame callback independently of continuous frame gating
- retain requests across empty or failed frames and roll back prepared callbacks when presentation fails
- cover both swapchain and direct software-pixel presentation paths
- gate callbacks from secondary windows without adding a window-list scan to the normal single-window path
- keep normal uncapped presentation immediately after the acknowledged frame

## Motivation

Continuous-rendering applications may disable global Wayland frame callbacks to preserve uncapped rendering. Some visual state transitions still need temporary compositor pacing. In a downstream game client, replacing an empty UI root with a character-selection UI could leave the new window invisible with callbacks disabled. Enabling callbacks fixed the transition but permanently capped rendering at the compositor refresh rate.

This API lets a caller synchronize only transition frames. The downstream integration requests one synchronized present per transition frame until the republished UI is included, then stops requesting them.

## Implementation

RequestPresentationSync latches an idempotent flag on the Context's active RenderTarget. The request survives an empty draw and is consumed by the next successful presentation.

The renderer uses an optional transactional PresentationSyncer before presentation. Wayland can prepare a callback even when GOGPU_WAYLAND_FRAME_CALLBACK=0; if swapchain or software presentation fails, the callback is canceled and a forced request is retained for retry. Other platforms keep their existing SyncFrame behavior.

Secondary callbacks set a renderer-level atomic marker. The app scans visible windows only while that marker is active, keeping the ordinary single-window path free of window-manager locking.

## Validation

- CGO_ENABLED=0 go test ./...
- go test -race ./...
- go vet ./...
- Linux, Windows/amd64, and macOS/arm64 builds
- presentation/frame-callback tests repeated 100 times
- steady-state sync microbenchmark: approximately 2.8-3.1 ns/op, 0 allocations
- visually verified with a v0.44.6 backport on Wayland: the UI transition is reliable and steady-state FPS remains uncapped

Local staticcheck reports only four pre-existing U1000 findings on main, unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **One-shot presentation synchronization** — `Context.RequestPresentationSync()` gates rendering after the next successfully presented frame without enabling continuous compositor pacing. Wayland supports forced `wl_surface.frame` callbacks for GPU and software presentation, including secondary windows; failed presentations cancel and retain the request for retry.
+
 ### Fixed
 
 - **macOS: `SetMenu` leaf items invisible on menu bar** (#449) — `applyMenu` added bare `NSMenuItems` to the main menu bar; macOS only renders items with submenus. Role-based leaf items (About, Quit, Preferences) are now routed to the App Menu submenu automatically. Non-role leaf items log a warning.
--- a/app.go
+++ b/app.go
@@ -1315,8 +1315,33 @@ func (a *App) Quit() {
 // On all other platforms, this always returns true.
 func (a *App) frameCallbackReady() bool {
 	if fg, ok := a.platWindow.(platform.FrameGater); ok {
-		return fg.FrameCallbackReady()
+		if !fg.FrameCallbackReady() {
+			return false
+		}
 	}
+
+	// Preserve the single-window fast path. Secondary windows are scanned only
+	// after one of them has actually prepared compositor frame gating.
+	if a.renderer == nil || !a.renderer.secondaryFrameGatePending.Load() {
+		return true
+	}
+	if a.windowManager == nil {
+		a.renderer.secondaryFrameGatePending.Store(false)
+		return true
+	}
+
+	a.windowManager.mu.RLock()
+	defer a.windowManager.mu.RUnlock()
+	for _, id := range a.windowManager.order {
+		w := a.windowManager.windows[id]
+		if w == nil || !w.visible || w.platWindow == nil {
+			continue
+		}
+		if fg, ok := w.platWindow.(platform.FrameGater); ok && !fg.FrameCallbackReady() {
+			return false
+		}
+	}
+	a.renderer.secondaryFrameGatePending.Store(false)
 	return true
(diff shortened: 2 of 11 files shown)


lkmavi:
## Corporate review (winit / Gio / Qt / Flutter frame-clock lens)

**Verdict: approve with notes** — solid one-shot compositor sync design; CI green; test coverage is enterprise-grade.

### Strengths
- Pre-present `wl_surface.frame` matches winit `pre_present_notify` / Gio / SDL3.
- Transactional prepare → present → cancel-on-failure with request restore is the right model for `GOGPU_WAYLAND_FRAME_CALLBACK=0` + rare UI transitions.
- Request survives empty frames; pixel path sync moved into `WriteSurfacePixels` (avoids double `SyncFrame` on the `pixelPresented` endFrame path).
- Secondary-window gate marker preserves the single-window fast path.
- `lockDisplay` around `PresentPixels` is a welcome Wayland correctness fix.

### Notes (non-blocking)
1. **`secondaryFrameGatePending` clear race (multi-window):** after a full “all ready” scan, `Store(false)` can overwrite a concurrent `Store(true)` from another window’s `PrepareFrameSync`, leaving an in-flight secondary callback invisible to the fast path. Unlikely on the single-threaded app loop; for strict multi-window correctness prefer storing “any still pending” after the scan (or CAS + re-check).
2. **Empty-frame pacing change:** continuous Wayland gating no longer `SyncFrame`s on empty frames. Correct for one-shot retention; worth a short CHANGELOG/docs note that empty-frame series may run uncapped when continuous callbacks are enabled.

Happy to re-review a follow-up if you harden (1). Thanks @kivutar — this unblocks the downstream transition use case cleanly.

kolkov (approved):
Second review — validated all claims from deep analysis against code. 20 tests + 1 benchmark, 0 blocking issues. Both concerns (atomic flag race window, PresentationSyncer naming) match @lkmavi's review and are non-blocking. Bonus: lockDisplay fix for WriteSurfacePixels. LGTM — @lkmavi please merge when ready.

kivutar:
Follow-up: the downstream issue this was intended to solve was ultimately traced to Goro mutating its character-selection widget tree from Draw, after gogpu/ui layout had completed. Incremental repaint then cleared the old window and painted an unlaid-out replacement. Moving the mutation into Update fixed the issue with vsync disabled. This Wayland change is therefore not needed for the original downstream bug.

kolkov:
Thanks for the follow-up — that kind of transparency is exactly what makes open source work well.

The change is architecturally sound regardless of the downstream trigger. One-shot compositor sync is a correct primitive for Wayland apps that render on-demand — it'll serve future use cases (theme transitions, power-aware rendering, compositor-paced popup reveals). The `lockDisplay` addition to `WriteSurfacePixels` also fixed a pre-existing race we hadn't caught. We're keeping it.

The downstream finding is interesting too — mutating the widget tree during Draw instead of Update is a subtle but real problem. Worth documenting for other consumers.

Welcome to the project, and looking forward to more contributions!

PR #429, opened 2026-08-06
release: v0.50.0 — outgoing drag-and-drop

kolkov:
## Summary

- **Outgoing drag-and-drop (drag source)** (#427, ADR-061) — `Window.StartDrag(DragData, callback)` on all 5 platforms
- **Windows DoDragDrop fixes** — correct HRESULT constant, OleInitialize ordering, IEnumFORMATETC, QueryInterface IID validation
- **macOS async drag callback** — `draggingSession:endedAtPoint:operation:` on GoGPUView
- **Documentation** — CHANGELOG, ROADMAP, ARCHITECTURE, README updated

## Test plan

- [x] Windows: drag file from window to desktop — COPIED (verified visually)
- [x] Build: Windows, Linux, macOS — clean
- [x] Lint: 0 issues on all 3 platforms
- [x] Tests: 8 packages pass
- [x] macOS: needs hardware verification (@jbunds or @lkmavi)
- [ ] Linux X11/Wayland: needs hardware verification (@unxed)

PR #453, opened 2026-08-13
fix(macOS): prevent NSRangeException on outgoing StartDrag

lkmavi:
## Summary

- Fixes macOS outgoing drag crash reported on #429: `NSDraggingItem.draggingFrame` was `{{0,0},{0,0}}`, AppKit raises `NSRangeException` and aborts the process during `beginDraggingSessionWithItems:`.
- Sets non-zero 32pt frame centered on the pointer (view coords via `convertPoint:fromView:`) plus `NSWorkspace.iconForFile:` preview before starting the session (JUCE / Qt / WebKit pattern).
- Hardening parity with Qt `qcocoadrag.mm`, JUCE `Windowing_mac.mm`, Flax: local `NSAutoreleasePool`, skip relative paths, one active outgoing drag per view.

## Test plan

- [x] `cd examples/drag_source && GOTOOLCHAIN=go1.25.0 go run .` — click-hold-drag to Desktop; no crash; file appears; console shows COPIED/MOVED/CANCELED
- [x] Relative-only paths — StartDrag completes with cancelled / no session (no AppKit exception)
- [x] Second StartDrag while first is in flight — rejected cleanly (no callback overwrite)
- [x] `GOTOOLCHAIN=go1.25.0 go test ./internal/platform/darwin/ -count=1`
- [x] `GOTOOLCHAIN=go1.25.0 go build ./...`
- [x] `golangci-lint run ./internal/platform/darwin/...`


kolkov (changes requested):
Thorough fix — enterprise references (Qt `qcocoadrag.mm`, JUCE `Windowing_mac.mm`, Flax) are properly cited and the hardening (autorelease pool, path filtering, re-entrancy guard) matches their patterns. The drag example rewrite is a big improvement.

One issue that needs a fix before merge:

### `ConvertPointFromView`: result buffer must be `[4]float64`

`var result [2]float64` at the end of `ConvertPointFromView` — goffi's `handleHFAReturn` always casts the result pointer to `*[4]float64` regardless of actual element count. Go 1.26's checkptr (enabled by `-race`) rejects smaller allocations. The existing `GetPoint` method already has this fix with an explicit comment (objc.go:920-923):

```go
// Result buffer sized to [4]float64 because goffi's handleHFAReturn always
// casts rvalue to *[4]float64 regardless of the actual element count, and
// Go 1.26's checkptr (enabled by -race) rejects a smaller allocation.
var result [4]float64
```

Change `[2]float64` → `[4]float64` in `ConvertPointFromView`.

### Minor observations (not blocking)

- `icon.SendSize` mutates the shared `NSWorkspace.iconForFile:` cached image. Qt/JUCE copy the icon first (`[icon copy]`). In practice macOS tends to return fresh copies, but a defensive copy would be safer.
- `SendRectPtr` returns `ID` but `setDraggingFrame:contents:` returns `void` — harmless since the return is discarded, but worth a doc note.

Everything else checks out: thread safety (mutex ordering, callback lifecycle, no double-release), error paths (all call `releaseDragSource`), FFI arg types, autorelease pool scope, test coverage.

lkmavi:
Addressed review feedback (@kolkov):

1. **Blocking:** `ConvertPointFromView` result buffer is now `[4]float64` with the same goffi `handleHFAReturn` / Go 1.26 checkptr comment as `GetPoint`.
2. **Non-blocking:** `fileDragIcon` copies via `[icon copy]` before `setSize:` and releases the owned copy after `setDraggingFrame:contents:` retains it.
3. **Non-blocking:** `SendRectPtr` docs note that `setDraggingFrame:contents:` is void and the `ID` return must be ignored.

Ready for re-review.

kolkov (approved):
Both issues addressed — `[4]float64` buffer for HFA return + defensive `[icon copy]` before `setSize:`. CI all green. LGTM.

Issue #449, opened 2026-08-11
fix(macOS): SetMenu leaf items invisible on menu bar + Role+Action guard


kolkov:
## Summary

`app.SetMenu()` adds menu items as bare NSMenuItems directly to the macOS main menu bar. macOS only renders menu bar items that have submenus — bare items are invisible. This makes `SetMenu` a silent no-op on macOS for leaf items.

Additionally, `addPlatformItem` skips the role path when both `Role` and `Action` are set (`item.Action == nil` guard), so custom actions on role items (About, Quit) don't work.

## Root Cause

Three related issues in `internal/platform/platform_darwin.go`:

**1. Leaf items invisible on menu bar** (`applyMenu`, line 1629)

`applyMenu` iterates user items and calls `addPlatformItem(mainMenu, item)`. For leaf items (no submenu, no role), this calls `AddMenuItemWithCallback` which adds a bare NSMenuItem to the main NSMenu. macOS menu bar only renders items with submenus.

`SetCustomMenu` works correctly because it wraps items in a submenu (`app.go:1691`). `GetSystemMenu().AddItem()` works because it adds to an existing App Menu submenu.

**2. Role + Action guard** (`addPlatformItem`, line 1659)

```go
if item.Role != MenuRoleNone && item.Action == nil {  // requires Action == nil
```

Items with both `Role` and `Action` skip the role path entirely. Note: `AddToSystemMenu` (line 1744) does NOT have this guard — inconsistency.

**3. API mismatch** (`SetMenu` vs `SetCustomMenu`)

`SetMenu` passes flat `MenuItem` list to `SetApplicationMenu`. `SetCustomMenu` correctly wraps each menu as `MenuItem{Title, Submenu}`. The flat path works on Windows (Win32 menu bar accepts leaf items) but not macOS.

## Reproduction

```go
app.OnSurfaceAvailable(func() {
    app.SetMenu(gogpu.NewMenu().
        AddItem(gogpu.MenuItem{Title: "About Foo", Action: func() { fmt.Println("about") }}).
        AddItem(gogpu.MenuItem{Separator: true}).
        AddItem(gogpu.MenuItem{Title: "Quit Foo", Role: gogpu.RoleQuit}))
})
// Result on macOS: default menu bar unchanged, custom items invisible
```

## Proposed Fix

1. **Role items → route to App Menu submenu.** Items with a Role (About, Quit, Preferences, etc.) should be added to the App Menu, not the menu bar root.

[...]

PR #457, opened 2026-08-14
fix(macOS): SetMenu replace, separators, Role+Action icons (#456)

### LINKED ISSUE #456: fix(macOS): SetMenu separators lost, system icon missing on Role+Action, append vs replace
## Follow-up to #449

Three remaining issues with `SetMenu` on macOS, reported by @jbunds in [discussion #423](https://github.com/gogpu/gogpu/discussions/423#discussioncomment-18001053) with screenshot.

### 1. Separators between Role items are invisible

`applyMenu` routes separators to `mainMenu` (the NSMenuBar) where they're invisible. When users interleave separators with Role items (e.g., About → separator → Quit), the separator goes to the bar while Role items go to the App Menu submenu.

**Root cause:** `platform_darwin.go` — condition `item.Separator || len(item.Submenu) > 0` always sends separators to `mainMenu`.

**Enterprise reference:** GLFW `cocoa_init.m:122,129,140` adds separators directly to `appMenu`, not the bar.

### 2. System icon (ⓘ) disappears when Role + Action specified

`AddMenuItemWithRole` uses `orderFrontStandardAboutPanel:` selector → AppKit renders system ⓘ icon. `AddMenuItemWithCallback` uses `handleMenuItem:` → plain NSMenuItem, no icon. Our #449 fix routes Role+Action to `AddMenuItemWithCallback`, losing the icon.

Screenshot confirms: "no callback" items have ⓘ, "with callback" items don't.

### 3. SetMenu appends to existing App Menu instead of replacing

`applyMenu` preserves the existing App Menu (index 0) with all system items (About, Settings, Hide, Quit). User Role items are *appended* after the existing system items, creating duplicates.

**Enterprise reference:** Electron `Menu.setApplicationMenu()` fully replaces the menu.

### Reproduction

```go
app.SetMenu(gogpu.NewMenu().
    AddItem(gogpu.MenuItem{Separator: true}).
    AddItem(gogpu.MenuItem{Title: "About Foo", Role: gogpu.RoleAbout}).
    AddItem(gogpu.MenuItem{Separator: true}).
    AddItem(gogpu.MenuItem{
        Title:  "Custom About",
        Role:   gogpu.RoleAbout,
        Action: func() { /* custom */ },
    }))
```

Expected: App Menu with separator, About (with icon), separator, Custom About (with icon).
Actual: Separators missing, Custom About has no icon, items appended after existing system items.

lkmavi:
## Summary

Follow-up to #449 / discussion #423 — enterprise-style `SetMenu` on macOS:

- **Separators** between Role items go to the App Menu (GLFW pattern); was routing them to the invisible menu bar
- **`separatorItem` singleton** — each separator is `copy`'d so multiple separators coexist (AppKit requirement)
- **Role+Action icons** — keep system selectors (`orderFrontStandardAboutPanel:` etc.) and override via `setTarget:` so AppKit still draws ⓘ / ⚙️ chrome (Electron/Qt pattern)
- **Replace semantics** — clear App Menu before applying caller items (`Menu.setApplicationMenu()` style), no more duplicate About/Hide/Quit
- **`RoleServices`** — create Services submenu + `setServicesMenu:` (was a no-op); default `createAppMenu` also attaches the missing Services item

Fixes #456

## Screenshot

`examples/menu` after fix:

SetMenu after #456

## Test plan

- [x] `go test ./internal/platform/ ./internal/platform/darwin/ -count=1`
- [x] Manual: `examples/menu` — App Menu shows separators, Services, system icons, Custom App Command; no duplicate system items
- [x] Role+Action About still shows ⓘ and fires custom `Action`
- [x] `GetSystemMenu(Application).AddItem(...)` still appends after `SetMenu` replace


kolkov (changes requested):
Excellent follow-up — all three #456 issues fixed cleanly. The `setTarget:` pattern for preserving role icons while wiring custom callbacks is the right enterprise approach (Electron/Qt). Separator `copy`, replace semantics, Services submenu — all validated against GLFW/AppKit docs. 6 new tests including the #456 reproduction scenario.

Two items to address:

### 1. Cache `copy` selector in `menuSels`

`newSeparatorItem` calls `RegisterSelector("copy")` inline on every invocation. Every other frequently-used selector in this file is cached in `menuSels`. For consistency:

```go
// in menuSels struct:
copy SEL

// in initMenuSelectors:
menuSels.copy = RegisterSelector("copy")
```

### 2. Clean `menuActionMap` entries on `removeAllItems`

With replace semantics, `SetMenu` can be called multiple times. Each call creates new items via `setMenuItemAction` (stores in `menuActionMap` keyed by item pointer). When `removeAllItems` removes the old items, their `menuActionMap` entries remain — stale `uintptr` keys pointing to deallocated NSMenuItems. If macOS reuses those addresses for new items, the stale callbacks could fire incorrectly.

Add a `clearMenuActions(menu)` helper that iterates the menu's items and calls `menuActionMap.Delete(uintptr(item))` for each, then call it before `removeAllItems`. Currently there is no deletion path for `menuActionMap` entries at all — only `Store` and `Load`.

This was pre-existing but is now more relevant since replace semantics encourages repeated `SetMenu` calls.

lkmavi:
Addressed both review notes in f5eb6ef:

1. `copy` is cached in `menuSels` (along with `numberOfItems` / `itemAtIndex:` / `submenu` used by the walker).
2. `ClearMenuActions` walks the menu tree (including nested submenus) and `Delete`s `menuActionMap` entries before `removeAllItems` on both the menu bar and App Menu.

`TestClearMenuActions_RemovesCallbacks` covers top-level + nested items.

kolkov (approved):
Both items addressed — `copy` selector cached in `menuSels`, `ClearMenuActions` with recursive submenu cleanup before `removeAllItems`. Test covers top-level + nested callbacks. CI all green. LGTM.

Issue #311, opened 2026-06-15
Enterprise API Audit — Public API Cleanup


kolkov:
## Enterprise API Audit — gogpu Public API Cleanup

Tracking issue for gogpu public API improvements. Sources: @lkmavi enterprise audit, 3 validation agents, 3 independent audit agents. Each item marked: ✅ verified, 💬 needs discussion.

### Architecture Context

gogpu is the **producer** in a shared GPU resource ecosystem. It creates the GPU device/queue and shares them with consumers (gg, ui, Born ML) via `gpucontext.DeviceProvider`. This is a **shared resource pattern** (like wgpu-rs → vello/bevy, or Go `database/sql` → drivers).

This differs from monolithic engines (Flutter, Unity) where the engine owns GPU resources internally. Our public API must serve ecosystem consumers who type-assert to `*wgpu.Device` for HAL access — the concrete + opaque two-layer design is intentional, not a leak.

### ✅ Unexport Internal Types (verified: zero external usage)

- [ ] `SurfaceState` + constants (`renderer.go:30`) — internal state machine
- [ ] `RenderTarget` (`renderer.go:42`) — internal per-window GPU state
- [ ] `WindowManager` (`window_manager.go:150`) — zero public methods
- [ ] `PlatformWindowCloser` (`window_manager.go:18`) — macOS delegate plumbing
- [ ] `AnimationController` (`animation.go:10`) — only used inside App.Run()
- [ ] `Invalidator` (`invalidator.go:8`) — only used inside App.Run()
- [ ] `GestureRecognizer` (`gesture.go:35`) — only used inside eventSourceAdapter

### ✅ Remove / Deprecate Dead Code

- [ ] `gogpu.DeviceProvider` interface (`device_provider.go:14`) — zero external usage, replaced by `gpucontext.DeviceProvider`. Deprecate `App.DeviceProvider()`, consumers use `App.GPUContextProvider()`
- [ ] `Context.Renderer()` (`context.go:168`) — leaks `*Renderer` internals. Only 1 usage: `examples/texture/main.go`. Alternative: `dc.AsTextureDrawer().TextureCreator()`. Deprecate accessor, update example
- [ ] Dead `dispatchMouseMove/Press/Release/Scroll/Resize` methods (`event_source.go:246-274`) — legacy, pointer events replaced them
[...]

PR #458, opened 2026-08-14
refactor: apply compatibility-safe API cleanup

besmpl:
## Summary

- replace exported file-dialog aliases with gogpu-owned types and explicit platform conversion
- remove dead event-source dispatch helpers and rename the surface-format filter helper
- deprecate legacy backend/provider entry points, update callers/docs, and correct DIP documentation

## Compatibility

The legacy `App.DeviceProvider`, `DeviceProvider`, `Config.Backend`, `Config.WithBackend`, and backend constants remain as source-compatible deprecated shims. Backend selection is controlled by wgpu build tags and `GraphicsAPI`; no unresolved Renderer/DrawTriangle/PresentTexture/Clear decisions are included.

Refs #311. This completes the mechanical/API-safe slice without closing the remaining design checklist.

## Verification

- `go test ./...`
- `go test -race ./...`
- `go build ./...`
- Linux, Windows, macOS, and js/wasm cross-builds
- `git diff --check`

`go vet ./...` reports only the repository's unchanged Darwin `unsafe.Pointer` diagnostics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,35 +212,18 @@ tex, err := renderer.LoadTextureWithOptions("tile.png", opts)
 
 ---
 
-## DeviceProvider Interface
+## GPU Resource Integration (gpucontext)
 
-GoGPU exposes GPU resources through the `DeviceProvider` interface for integration with external libraries:
-
-```go
-type DeviceProvider interface {
-    Device() hal.Device              // HAL GPU device (type-safe Go interface)
-    Queue() hal.Queue                // HAL command queue
-    SurfaceFormat() gputypes.TextureFormat
-}
-
-// Usage
-provider := app.DeviceProvider()
-device := provider.Device()   // hal.Device — 30+ methods with error returns
-queue := provider.Queue()     // hal.Queue — Submit, WriteBuffer, ReadBuffer
-```
-
-### Cross-Package Integration (gpucontext)
-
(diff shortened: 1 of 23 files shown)


kolkov (commented):
The code changes are solid — `FileDialogOptions` with proper deep-copy, dead dispatch helpers correctly removed, `doc.go` rewrite fixes real factual errors ("macOS: Planned" was wrong since v0.21.0), and the `allowedSurfaceFormat` rename improves clarity.

**Question for discussion — deprecated shims pre-v1.0:**

This PR adds `// Deprecated:` annotations and keeps backward-compatible shims for `Config.Backend`, `WithBackend()`, `DeviceProvider`, and the backend constants. All internal callers are migrated — the shims serve hypothetical external consumers.

Our pre-v1.0 rule is "do the RIGHT architecture immediately, no phased compromises." At v0.53.0 with no stable API contract, should we delete the deprecated API entirely instead of maintaining shims? Or is there a specific external consumer we're preserving compatibility for?

Options:
1. **Delete now** — clean API surface, no maintenance burden, pre-v1.0 is the time
2. **Keep shims** — if there's a known consumer (Born ML, KiGo, etc.) that uses `DeviceProvider()`
3. **Delete in follow-up PR** — merge this as-is, delete in next release

What's your reasoning for the shim approach, @besmpl?

**Minor note:** The `OnResize` test now only checks callback registration (`!= nil`) instead of dispatch. The removed `dispatchResize` was dead code (never called from app.go — real dispatch tested in `app_coverage_test.go`), so the coverage impact is narrow.

Issue #243, opened 2026-05-16
feature request: support native printing dialog boxes


celer:
I want to be able to use the gogpu/ui to open a native printing dialog to print PDFs, etc.  I can't do this today with the framework I'm using this probably requires some integration with gogpu/gg etc.

besmpl:
Implemented and merged via #459.

The merged work adds the public print contract plus native Windows, macOS, and Linux backends with deterministic syscall/bridge tests, cancellation/lifecycle handling, platform cross-build coverage, and documentation of the Windows raw-PDF PASSTHROUGH driver limitation. CI and Codecov were green at merge.

kolkov:
Implemented in v0.53.0 via PR #459 (@besmpl) — native desktop print backends for Windows, macOS, and Linux. Public API: `App.Print()`, `App.PrintWithOptions()`. See `docs/PRINTING.md` for usage.

Thanks for the feature request, @celer! If you run into any issues with the printing API, please open a new issue.

PR #459, opened 2026-08-14
feat: add native desktop print backends

besmpl:
## Summary
- keep the backend-neutral `App.Print` contract for caller-supplied complete documents
- add Windows `PrintDlgExW`/GDI spooling with parent HWND, ranges, copies, and cancellation
- add macOS PDFKit/AppKit printing with main-thread panel dispatch, parent window ownership, ranges, copies, and cancellation
- add Linux X11/Wayland xdg-desktop-portal PreparePrint/Print flow with parent tokens, temporary document FD lifetime, ranges, copies, and cancellation
- unify native job lifecycle/cancellation semantics and map unavailable native services to `ErrPrintUnsupported`

## Verification
- `gofmt -l .`, `git diff --check`
- `go test -count=1 ./...`, `go test -race -count=1 ./...`, `go build ./...`
- Windows amd64/arm64 builds and platform test compilation
- macOS amd64/arm64 builds and platform test compilation
- Linux amd64/arm64 builds and platform test compilation
- Linux platform vet passes
- `golangci-lint` runs with only two pre-existing `goconst` findings (`internal/platform/darwin/menu.go`, `sound/sound_darwin.go`)

## Validation limits
- interactive Windows printer/dialog, macOS AppKit panel, and Linux portal/spool runs require their native hosts and were not available here
- Windows forwards the caller's complete document through GDI `RAW`; printer-driver support for PDF/document input remains a native-environment concern
- browser/WASM printing remains unsupported

Refs #243

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Built on [gogpu/wgpu](https://github.com/gogpu/wgpu) — the unified Go WebGPU p
 | **Platforms** | Windows (Vulkan/DX12/GLES), Linux X11/Wayland (Vulkan/GLES), macOS (Metal), Browser/WASM (WebGPU) |
 | **Rendering** | Event-driven three-state model (idle/animating/continuous), 0% CPU when idle (lazy acquire), zero-copy surface rendering, damage-aware presentation |
 | **Input Models** | Three coexisting models: callbacks (`EventSource`), state polling (`Input()`, Ebiten-style), SDL-style event queue (`PollInputEvent()`) |
-| **Graphics** | Windowing, input handling, multi-keyboard layout (X11 XKB + Wayland xkbcommon), AltGr/international text input (unified xkbcommon, ADR-029), key repeat on all platforms (Wayland client-side timer, ADR-033), texture loading, frameless windows, runtime window resize (`RequestSize`), mouse grab / pointer lock (Win32 + X11 + Wayland, SDL parity), OS drag-and-drop: incoming (all 4 desktop platforms) + outgoing drag source (`StartDrag`, all 5 platforms), GPU adapter power preference, native macOS window tabbing, native system menus (macOS + Windows), native file dialogs (macOS + Windows + Linux D-Bus/zenity/kdialog) |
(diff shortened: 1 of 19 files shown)


kolkov (changes requested):
Strong PR — well-layered architecture, 34 tests with syscall seam for deterministic testing, zero CGO on all platforms. The print job state machine is textbook concurrent lifecycle management. Lint issues already fixed in latest push. CI all green.

### Must fix

**1. `nStartPage` must be `START_PAGE_GENERAL` (0xFFFFFFFF), not zero.**

The `printDlgEx` struct literal does not set `nStartPage`, so it defaults to 0. Per [MSDN PRINTDLGEXW](https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-printdlgexw), `nStartPage` must be `START_PAGE_GENERAL` (0xFFFFFFFF) to show the General tab. Zero means "first property page" which can show wrong UI on some printer drivers.

```go
dlg := printDlgEx{
    // ...existing fields...
    nStartPage: 0xFFFFFFFF, // START_PAGE_GENERAL
}
```

**2. Document PASSTHROUGH limitation in `docs/PRINTING.md`.**

The Windows backend sends raw PDF bytes via GDI `PASSTHROUGH` escape. This only works on drivers supporting raw document formats (PostScript, PCL, virtual PDF printers like "Microsoft Print to PDF"). Consumer inkjet printers typically do not support PASSTHROUGH and will return an error. The doc currently says "Windows may render PDF before submitting" which is inaccurate — it sends raw bytes, no rendering occurs. Add a Windows-specific note.

### Non-blocking observations

- **Triple PDF copy on Windows** — `NewPDFDocument` + `App.Print` + `StartPrint` copies data 3 times. macOS/Linux skip the third copy. Consider removing the redundant copy in `StartPrint`.
- **Struct alignment** — verified: `sizeof=136`, all offsets match MSDN x64. Adding `unsafe.Offsetof` assertions in a test would prevent regressions.
- **D-Bus wait deadline** — `waitResponseBody` 5-min hardcoded deadline doesn't use `min(5min, ctx.Deadline())`. Context cancellation still works via socket close.

Everything else checks out: macOS main-thread dispatch, PDFKit page filtering, Linux FD passing via SCM_RIGHTS, secure temp files (0600), cancellation semantics.

kolkov (approved):
nStartPage fixed with constant + test assertion. CI all green. PASSTHROUGH docs — we'll add a Windows limitations note ourselves. LGTM.

PR #461, opened 2026-08-14
perf: reuse frame encoder and batch textured quads

### LINKED ISSUE #329: perf(renderer): batch textured-quad blits into instanced draws (gpui-style)
## Context

Studying Zed's **gpui** rendering for techniques applicable to this ecosystem. gpui's core throughput trick is **instanced quad rendering**: it collects all quads of a kind into one buffer and issues a *single* instanced draw, rather than one draw call per quad. This stack has already adopted most of gpui's playbook (glyph atlas, SDF shapes, 0-alloc hot paths, glyph-quad batching via `buildGlyphMaskDrawCalls`), but the **textured-quad** path is still one-draw-per-quad.

## Observation (grounded in source)

`drawTexturedQuad` (renderer.go:991) renders **one quad per call**: it writes the quad's params (X, Y, W, H, surface dims, alpha) into a **single shared uniform buffer** `texQuadUniformData` (renderer.go:1037-1043) and issues a draw. Because the uniform is overwritten each call, multiple quads cannot share a draw.

The main consumer is the compositor in **gogpu/ui** (`desktop/desktop.go` `compositeTexturesFromTree`, ~line 727), which walks the layer tree and blits **one `DrawGPUTexture*` call per repaint-boundary PictureLayer** (desktop.go:793 `DrawGPUTextureBase`, 802-805 `DrawGPUTextureWithOpacity`). A UI with N concurrently-dirty boundaries therefore issues **N textured-quad draw calls + N uniform writes per frame** to composite.

For static UIs this is moot (frame-skip keeps GPU at 0% per the ui ROADMAP), and N is usually small — so this matters specifically for **complex/animated UIs with many simultaneous boundary blits** (dashboards, IDE layouts, lots of independently-animating subtrees).

## Proposal (gpui-style instanced textured quads)

Add a batched textured-quad path alongside the per-quad one:

1. **Per-instance data** — replace the single shared uniform with a per-instance storage/instance buffer: an array of `{x, y, w, h, alpha, atlasOrTexIndex, ...}`. Surface dims go in a frame-global uniform.
2. **Batch key** — quads sharing the same source texture (or atlas page) and pipeline state batch into one instanced draw (`draw(6, instanceCount)` or indexed). Distinct textures need distinct bind groups; a texture-array / atlas would collapse those too.
3. **Batched API** — expose something like `DrawTexturedQuads([]TexturedQuad)` (or a builder that accumulates and flushes once per pass), mirroring the existing glyph-mask `GlyphMaskBatch` → `buildGlyphMaskDrawCalls` design that already lives in `internal/gpu`.
4. **Consumer** — gogpu/ui's `compositeTexturesFromTree` accumulates all boundary blits for a pass and submits them as one batch instead of N immediate `DrawGPUTexture` calls. Clip/opacity variants become per-instance fields.

This reuses the exact pattern already proven for glyph masks in this repo, just for compositor/image quads.

## Expected benefit

Fewer draw calls and uniform writes per frame, scaling with boundary count. Biggest win on UIs that composite many dirty boundaries per frame.

## Honesty / validation note

I can't benchmark this myself — I don't have a GPU/display in my current environment, so I'm filing this as a **design proposal** rather than a PR with measured numbers. Before implementing, this should be validated on real GPU hardware (draw-call count + frame time on a many-boundary scene, e.g. the `examples/ide` or a synthetic N-boundary stress test). Happy to prototype the CPU-side batching/API and the ui compositor consumer if a maintainer can run the GPU-side measurements, or to scope it down to the atlas/texture-array question first.

Refs: `renderer.go:991-1043` (per-quad uniform), gogpu/ui `desktop/desktop.go:727,793-805` (per-boundary blit loop), `internal/gpu` glyph-mask batching as the template.

besmpl:
## Summary

- Reuse the frame-owned command encoder for textured-quad draws and submit it once at frame end.
- Keep one render pass open for each contiguous run of textured quads, while retaining per-draw dynamic uniform payloads.
- Preserve ordering and load semantics across external `Context.CommandEncoder` use, explicit clears, composition blits, debug overlays, standalone triangle draws, `RenderToImage`, and frame submission.
- Keep uniform writes deferred until the pass closes so failed-pass and frame-lifetime guarantees remain intact.

## Scope

This is the evidence-backed P1/P2 performance fix for #329. It changes private encoder/pass lifetime and batching only; there are no public API, storage-buffer, texture-atlas, or same-texture instancing changes. A broader instancing design remains deferred until a concrete workload and maintainer-approved contract justify it.

## Verification

- Focused textured-quad, encoder-boundary, clear-boundary, and lifetime tests
- `go test . -count=1`
- `go test ./... -count=1`
- `go test . -race -count=1`
- `go test -tags nogpu ./... -count=1`
- Host/Linux/Windows/JS-WASM builds
- `golangci-lint` v2.12.2 (`config verify`, `run --timeout=5m`)
- Benchmark confirms one encoder, one pass, and one submit for 1/20/100 contiguous draws

Closes #329.

### DIFF
--- a/context.go
+++ b/context.go
@@ -297,6 +297,13 @@ func (c *Context) CommandEncoder() *wgpu.CommandEncoder {
 		slog.Error("gogpu: shared frame encoder unavailable", "err", err)
 		return nil
 	}
+	// A textured-quad run may keep its render pass open for batching. Close it
+	// before lending the encoder to an external renderer so its pass can be
+	// recorded in the exact call-order position requested by the caller.
+	if err := ws.endTexturedQuadPass(); err != nil {
+		slog.Error("gogpu: shared frame encoder unavailable after textured quad pass", "err", err)
+		return nil
+	}
 	// Deferred clears must precede every externally recorded pass. Flushing
 	// here preserves call order; waiting until EndFrame would clear overlays.
 	if !ws.flushClear(ws.renderer.device, ws.renderer) {
--- a/context_command_encoder_test.go
+++ b/context_command_encoder_test.go
@@ -157,6 +157,20 @@ func TestContextCommandEncoderReusesAndSubmitsFrameEncoderOnce(t *testing.T) {
 	}
 }
 
+func TestBeginFrameReleasesPreviousTexturedQuadUniformArena(t *testing.T) {
+	ws := &RenderTarget{
+		texQuadUniformChunks: []texQuadUniformChunk{{nextSlot: 7}},
+	}
+	r := &Renderer{primary: ws}
+
+	if r.BeginFrame() {
+		t.Fatal("BeginFrame succeeded for an unconfigured surface")
+	}
+	if got := len(ws.texQuadUniformChunks); got != 0 {
+		t.Fatalf("textured-quad uniform chunks after new frame = %d, want 0", got)
+	}
+}
+
 func TestDrawTexturedQuadsReuseFrameEncoderAndPreserveLoadOrder(t *testing.T) {
 	halDevice := &recordingCommandEncoderDevice{}
 	ctx, target, queue := newSharedEncoderTestContextWithDevice(t, halDevice)
--- a/context_command_encoder_test.go
+++ b/context_command_encoder_test.go
@@ -13,7 +13,8 @@ import (
 
 type countingSubmitQueue struct {
 	noop.Queue
-	submits int
+	submits        int
+	writeBufferErr error
 }
 
 type commandEncoderFailDevice struct{ noop.Device }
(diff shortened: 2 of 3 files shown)


kolkov (approved):
Good work on the encoder reuse — the architecture matches enterprise references (GTK4 GSK record-then-merge, Flutter Impeller 1-encoder-per-frame). The boundary handling is thorough, and the benchmarks confirm the optimization works (N encoders + N submits → 1+1).

**One fix needed before merge:**

`abortTexturedQuadPassState` silently discards the `pass.End()` error:

```go
_ = ws.texturedQuadPass.End()
```

Per our error handling rules, this should log the error:

```go
if err := ws.texturedQuadPass.End(); err != nil {
    slog.Warn("gogpu: end textured quad pass during abort", "err", err)
}
```

It's an abort path so non-fatal, but silent discard hides diagnostic info.

**Minor suggestions (non-blocking):**

- The benchmark exercises 100 draws (>64 slots), but there's no dedicated test asserting second-chunk allocation correctness. Consider adding one.
- `testContextHelper` interface could be `testing.TB` — it's a strict subset.

After the `_ =` fix, this is ready to merge.

Issue #331, opened 2026-06-23
feat: Input Method Editor (IME) support — CJK text input across all platforms


kolkov:
## Summary

Full Input Method Editor (IME) support for CJK (Chinese/Japanese/Korean) text input, emoji pickers, dead key sequences, and voice dictation across all platforms.

**Related:** [ui#144](https://github.com/gogpu/ui/issues/144) (@diiyw) — original feature request

## Motivation

CJK users (~1.5B people) require IME for text input. Phonetic input (e.g., Pinyin "nihao" → 你好) requires a **composition preview** (preedit) that shows intermediate text before the final ideographic selection. Without IME support, CJK users cannot use TextField or any text input widget.

Currently gogpu only delivers **committed text** via `WM_CHAR` (Windows), `[NSEvent characters]` (macOS), and `xkb_state_key_get_utf8` (Linux). There is no composition preview, no candidate window positioning, and no content type hints.

## Architecture (ADR-045)

Following the **winit** pattern (closest architectural match — windowing library, not full framework):

- **gpucontext**: Extended `IMEController` interface (cursor area, content type, surrounding text, cancel) + `IMEEventSource` for disabled/delete-surrounding events + `ContentPurpose` (15 values) / `ContentHint` (10 flags) types
- **gogpu**: Platform-specific IME integration per OS
- **ui**: TextField composition rendering (preedit underline, cursor within composition)

### Platform APIs

| Platform | API | Model |
|----------|-----|-------|
| Windows | IMM32 (`WM_IME_*` messages + `imm32.dll`) | Push |
| Wayland | `zwp_text_input_v3` protocol | Push (atomic commit) |
| macOS | `NSTextInputClient` protocol (11 methods) | Pull (system queries app) |
| X11 | XIM (`XOpenIM`, `XCreateIC`, `Xutf8LookupString`) | Push |
| Browser | Hidden `<input>` + `CompositionEvent` DOM events | Push |

### Key Design Decisions

- **IME disabled by default** per window (winit pattern) — widgets enable on focus
- **Double-dispatch prevention** on every platform (IME path vs direct text path must not both produce events)
- **IMM32 only** on Windows (TSF not needed — SDL3, GLFW, winit all use IMM32)
- **IBus/Fcitx** on Wayland work through compositor automatically (zero D-Bus integration)
- **`IMEState.CursorBegin/CursorEnd`** — cursor range, not single point (both platforms provide range)

## Phased Implementation

[...]

PR #462, opened 2026-08-14
feat: add cross-platform IME backends

besmpl:
## Summary

- add a backend-neutral IME event/controller path with UTF-8 byte ranges and exactly-once committed-text routing
- implement Windows IMM32, Wayland text-input-v3, X11 XIM/XIC, macOS NSTextInputClient, and browser/WASM composition backends
- preserve focus/enable/cancel ordering, cursor areas, surrounding-text privacy, content hints, native teardown, and legacy fallback behavior
- harden password/sensitive sessions, late native callbacks, browser/Windows echo deduplication, and X11 key-release forwarding

## Dependency

This PR depends on [gpucontext #30](https://github.com/gogpu/gpucontext/pull/30), currently at `30ff593`, which adds the versioned optional IME interfaces.

The branch intentionally keeps the released `github.com/gogpu/gpucontext v0.28.0` requirement and contains no committed `replace`. It therefore remains draft until #30 is merged and released. After that release, update `go.mod`, run `go mod tidy`, rerun the full matrix, and mark this PR ready.

## Verification

All dependency validation used a temporary local workspace/replace that was removed before commit.

- focused IME/browser/platform regression tests: PASS
- `go test -count=1 ./...`: PASS
- `go test -race -count=1 ./...`: PASS
- supported CGO-disabled builds and platform test compilation for Linux amd64/arm64, Windows amd64/arm64, Darwin amd64/arm64, and js/wasm: PASS
- Linux X11 amd64/arm64 test compilation: PASS
- `gofmt` and `git diff --check`: PASS

Host vet reports only existing Darwin unsafe-pointer diagnostics. Live Windows IME, Wayland compositor, XIM locale, AppKit, and browser-engine validation requires target hosts and remains a post-merge/release validation item.

Refs #331. Depends on gogpu/gpucontext#30.


kolkov (commented):
The IME architecture is strong — clean three-layer design, comprehensive privacy model (centralized policy, auto-disable for passwords, teardown safety via `imeDestroyed` flags), and thorough test coverage for all platform state machines.

**Blocking: gpucontext #30 dependency**

This PR depends on gpucontext #30 which has open review items (see [our review](https://github.com/gogpu/gpucontext/pull/30#pullrequestreview-4949847331)). CI cannot pass until #30 is resolved and released. The 6 items from that review still need discussion — particularly the V2 interface pattern vs ADR-045.

**Code issues to address (after #30 resolves):**

1. **`_ = h.flushWithRetry()` (3 occurrences in Wayland)** — the 4th call in the same file properly checks the error, proving it's meaningful. These 3 should at minimum use `slog.Warn`.

2. **13 `_, _ = ffi.CallFunction(...)` in X11/macOS** — per our error handling rules, these need at least a package-level doc comment explaining why the errors are non-recoverable after nil-check preconditions.

3. **`x11IMEContentIsSensitive` duplicates centralized policy** — the centralized `imeContentIsSensitive` in `ime_privacy.go` is unexported, so X11 sub-package can't call it. Fix: export it as `IMEContentIsSensitive` in `internal/platform/`, or extract to a shared sub-package like `internal/platform/imepolicy/`.

4. **`validIMECursorArea` duplicated 5 times** (3 named functions + 2 inline checks). Same core logic (>=0, !NaN, !Inf). Wayland adds MaxInt32 bound. Consider a shared function with optional bounds parameter.

**What's done right (keep as-is):**
- UTF-8 byte offsets matching Wayland/GTK4/winit consensus
- `ContentPurposeChat` extension beyond protocol (Flutter pattern)
- Per-platform `defaultIMECapabilities()` with accurate feature flags
- Centralized `imeContentIsSensitive` policy design (just needs export)
- All 5 backends consistently auto-disable IME for sensitive content
- `imeDestroyed`/`closed` flags guard every public method against late callbacks

Waiting for gpucontext #30 resolution before this can proceed.

PR #466, opened 2026-08-15
fix(wayland): avoid eventMu reentrancy deadlock on scale change (#448)

### LINKED ISSUE #448: bug: Wayland window freeze on multi-display move (fractional scaling deadlock)
## Problem

Moving a window between displays with different fractional scaling causes the application to freeze ("Not Responding"). The render thread deadlocks during swapchain reconfiguration triggered by `wp_fractional_scale.preferred_scale` change.

**Reported by @porjo** in [g3d#32](https://github.com/gogpu/g3d/issues/32). Affects all gogpu apps (confirmed on both `g3d/examples/hello-cube` and `ui/examples/gallery`).

### Environment
- Fedora 44, Macbook Pro M2 (Asahi Linux)
- External monitor via USB-C (different fractional scale than laptop display)
- KDE Plasma Wayland
- Setting both displays to 100% scale → problem disappears

### Root Cause Analysis

@porjo captured `WAYLAND_DEBUG=1` output ([log](https://github.com/gogpu/g3d/issues/32#issuecomment-5249642453)) confirming:

> The hang occurs during (or immediately after) the Vulkan/Mesa surface reconfiguration triggered by the fractional-scale change that accompanies a monitor transition. The render thread (or the shared Wayland display lock used by both the main thread and Mesa's Vulkan WSI) appears to get stuck while rebuilding the swapchain / re-importing dmabufs for the new scale/output.

Our investigation found two contributing factors:

1. **Frame callback deadlock on output change.** The frame callback state machine (`libwayland_frame.go`) gates rendering — state `FrameCallbackRequested` blocks all frames. During output transition, the compositor may not deliver `wl_callback.done`, leaving the state machine stuck permanently.

2. **`displayMu` lock contention.** `DispatchDefaultQueue()` holds `displayMu` during the entire dispatch cycle. The render thread also acquires `displayMu` for `vkAcquireNextImageKHR` and `Present`. A fractional-scale change that forces swapchain reconfigure creates extended lock hold times, deadlocking the event loop.

3. **Missing `wl_surface.enter/leave` handlers on C surface.** The C surface (`LibwaylandHandle`) never registers listeners for `wl_surface.enter`/`wl_surface.leave` events. When the surface moves to a new output, no code updates the output reference, scale factor, or resets the frame callback.

### Related

- #409 (DPI/scale factor change — ADR-059) — same underlying gap (Wayland output change not handled)
- [g3d#32](https://github.com/gogpu/g3d/issues/32) — original report

### Proposed Fix

1. Register `wl_surface.enter/leave` listeners on C surface → update output + scale + reset frame callback
2. Add frame callback timeout recovery: if `FrameCallbackRequested` > 500ms without `done` → reset to `None`
3. Narrow `displayMu` lock scope during reconfigure to prevent deadlock with render thread

### Files

| File | What to change |
|------|---------------|
| `internal/platform/wayland/libwayland_xdg.go` | Register `wl_surface` enter/leave on C surface |
| `internal/platform/wayland/libwayland_frame.go` | Add timeout recovery for stuck frame callback |
| `internal/platform/platform_linux.go` | Handle output change — update scale, trigger reconfigure |
| `internal/platform/wayland/libwayland.go` | Track current output on `LibwaylandHandle` |

lkmavi:
## Summary
- Fixes window freeze when moving between Wayland outputs with different fractional scales (#448).
- Root cause from porjo's goroutine dump: `PrepareFrame` held `eventMu` then called `LogicalSize()`, which locks the same non-reentrant mutex — self-deadlock on the render thread after `preferred_scale` change.
- Reads logical size under the existing lock instead, and adds enterprise-style regression coverage (timeout deadlock check, secondary window, round-trip 1.0↔1.667, concurrent `LogicalSize`, env scale override).

## Test plan
- [x] `GOOS=linux GOARCH=arm64 go test -c ./internal/platform` then run `TestWaylandPrepareFrame*` in linux/arm64 container — all 7 PASS
- [x] Reproduce on dual-display KDE Wayland (fractional scales differ): drag window between outputs — app must stay responsive
- [x] Confirm `EventScaleChanged` still fires with correct logical size on scale transition

Closes #448

### DIFF
--- a/internal/platform/platform_linux.go
+++ b/internal/platform/platform_linux.go
@@ -793,7 +793,11 @@ func (w *waylandPlatformWindow) PrepareFrame() PrepareFrameResult {
 	wp.eventMu.Lock()
 	scaleChanged := wp.lastScale != 0 && wp.lastScale != scale
 	if scaleChanged {
-		lw, lh := w.LogicalSize()
+		// Read size under the same lock. Do NOT call LogicalSize() here:
+		// LogicalSize also acquires eventMu, and sync.Mutex is not reentrant —
+		// that self-deadlocks the render thread on multi-display fractional
+		// scale transitions (issue #448; confirmed via goroutine dump).
+		lw, lh := wp.width, wp.height
 		wp.events.Push(Event{
 			WindowID:    wp.winID,
 			Type:        EventScaleChanged,
--- a/internal/platform/platform_linux_prepare_frame_test.go
+++ b/internal/platform/platform_linux_prepare_frame_test.go
@@ -1,0 +1,262 @@
+//go:build linux
+
+package platform
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/gogpu/gogpu/internal/platform/eventqueue"
+)
+
+// newWaylandPrepareFrameFixture builds a minimal primary-window setup that
+// exercises PrepareFrame scale-change detection without a live compositor.
+func newWaylandPrepareFrameFixture(width, height int, outputScale, lastScale float64) (*waylandPlatformWindow, *waylandWindow) {
+	wp := &waylandWindow{
+		width:     width,
+		height:    height,
+		lastScale: lastScale,
+		winID:     1,
+		events:    eventqueue.New[Event](eventqueue.DefaultCapacity),
+		startTime: time.Now(),
+	}
+	plat := &waylandPlatform{
+		primary:     wp,
+		outputScale: outputScale,
+	}
+	return &waylandPlatformWindow{platform: plat, id: 1}, wp
+}
+
+// newWaylandSecondaryPrepareFrameFixture mirrors the primary fixture on the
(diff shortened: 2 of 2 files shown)


kolkov (approved):
One-line fix for a self-deadlock — `PrepareFrame` held `eventMu` then called `LogicalSize()` which locks the same non-reentrant mutex. Reading `wp.width, wp.height` directly under the held lock is correct and clean.

7 enterprise-grade regression tests: deadlock timeout check, secondary window, first-frame guard, steady-state no-spam, round-trip 1.0↔1.667 transitions, concurrent LogicalSize stress, env scale override. All edge cases from the #448 investigation covered.

Our ADR-071 research suspected `displayMu` contention — the real root cause (goroutine dump from @porjo) was simpler and more precise. Good lesson: reproduce first, theorize second.

LGTM.

PR #470, opened 2026-08-18
docs: fix broken star history chart in README

OctoBored:
The Star History chart in the README is currently broken and no longer renders for visitors, which makes the repository look less established than it is. This change points the chart and its wrapping link to a working endpoint so the chart displays correctly again.

No behavior beyond the chart link is affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -802,11 +802,11 @@ MIT License — see [LICENSE](LICENSE) for details.
 
 ## Star History
 
-<a href="https://star-history.com/#gogpu/gogpu&Date">
+<a href="https://star-history.dera.page/#gogpu/gogpu&Date">
  <picture>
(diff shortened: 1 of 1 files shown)


kolkov:
Thanks! Same issue was independently reported in ui#231 by @CrustyMozarella — star-history.com has been down (503) since GitHub restricted the stargazers API.

We're building **[starhistory.io](https://starhistory.io)** — our own service powered by gogpu/gg for chart rendering. It will replace all third-party charts across the ecosystem.

Merging this as a quick fix, will update to starhistory.io when it launches. Appreciate the contribution!

OctoBored:
You are too late tbh

kolkov:
We initially planned to merge this as a quick fix, but on reflection it makes more sense to skip the intermediate step — [starhistory.io](https://starhistory.io) is close to launch and will replace all third-party chart URLs across the ecosystem in one pass. Merging a temporary redirect now would just mean another README update soon after.

Closing in favor of the starhistory.io rollout. Thanks for flagging the broken chart, @OctoBored!

Issue #464, opened 2026-08-14
`examples/menu` does not render a Window menu


jbunds:
`examples/menu` does not render a **Window** menu in the menu bar.

kolkov:
Thanks for the report, @jbunds. This is a confirmed bug — same root cause as #463 (Cmd+W not working on About panel).

**Root cause:** `applyMenu()` in `platform_darwin.go` calls `removeAllItems` on the main NSMenu, then re-adds only the App Menu item. The Window menu (created by `createMenuBar()` during init with Minimize, Zoom, Close/Cmd+W, Full Screen, Bring All to Front) is removed and never re-added.

We checked Qt6, GLFW, Flutter, and Electron — all enterprise frameworks auto-preserve the Window menu:
- **GLFW**: unconditionally creates Window menu with Minimize, Zoom, Full Screen, Bring All to Front + `setWindowsMenu:`
- **Qt6**: auto-creates a hidden Window menu + `setWindowsMenu:` (even if user doesn't ask for one)
- **Flutter**: Window menu in MainMenu.xib template
- **Electron**: auto-adds Window menu items if user includes `role: 'windowMenu'`

**Fix:** preserve the Window menu NSMenuItem alongside the App Menu before `removeAllItems`, then re-insert after user items. This is @lkmavi's macOS domain — assigning.

Issue #463, opened 2026-08-14
selecting a custom `RoleAbout` menu item from an application menu set via `SetMenu` ignores `⌘+w` on macOS


jbunds:
Compare the behavior of the `examples/window_header` program versus `examples/menu`:

Selecting the **About GoGPU** application menu item when running the `examples/window_header` program and then entering `⌘+w` closes the window that pops up.

But selecting the **About My App** application menu item when running the `examples/menu` program ignores `⌘+w` even when the window that pops up has focus.

kolkov:
Thanks for the report, @jbunds. This is the same root cause as #464 (missing Window menu).

**Root cause:** `applyMenu()` removes the Window menu from the menu bar. The Window menu contains the "Close" item with `keyEquivalent: "w"` and `performClose:` action. Without it, Cmd+W has no menu item to route through — the key equivalent is dead.

In `examples/window_header` (default menu), the Window menu is intact → Cmd+W works. In `examples/menu` (custom `SetMenu`), `applyMenu()` strips the Window menu → Cmd+W stops working.

The About panel itself is not special — `orderFrontStandardAboutPanel:` creates a standard NSPanel. Cmd+W works via the same `performClose:` responder chain as any window. It needs a "Close" menu item with `keyEquivalent: "w"` visible in the menu bar to fire.

**Fix:** same as #464 — preserve the Window menu across `SetMenu` calls. Enterprise references (Qt6, GLFW, Flutter, Electron) all auto-preserve it.

PR #479, opened 2026-08-25
fix(macOS): preserve Window menu across SetMenu (#463, #464)

lkmavi:
## Summary
- `applyMenu` no longer drops the system Window menu (Minimize, Zoom, Close/Cmd+W, Full Screen, Bring All to Front) when replacing the menu bar via `SetMenu`
- Window menu item is detached + retained before `removeAllItems`, then re-inserted after user items (GLFW / Qt6 / Electron pattern)
- If the bar was already stripped, wraps `NSApp.windowsMenu` in a new item so recovery still works
- Fixes missing Window menu in `examples/menu` (#464) and dead Cmd+W on About panel after custom `SetMenu` (#463)

## Test plan
- [x] `go test ./internal/platform/ -run 'MacOSMenu|FindWindows'`
- [x] Run `examples/menu` on macOS: Window menu visible in menu bar
- [x] About panel → ⌘W closes the panel
- [x] `examples/window_header` still has working ⌘W (default menu path)
- [x] Call `SetMenu` twice in a row — Window menu remains, no duplicate crash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **macOS: Window menu lost after `SetMenu`** (#463, #464) — `applyMenu` preserved only the App Menu item, so Minimize/Zoom/Close (Cmd+W)/Full Screen disappeared from the menu bar. The system Window menu is now detached, retained, and re-inserted after user items (GLFW/Qt6/Electron pattern).
 - **macOS: `SetMenu` separators lost between Role items** (#456) — top-level separators were routed to the menu bar (invisible). They now go to the App Menu submenu, matching GLFW `cocoa_init.m`.
 - **macOS: About ⓘ icon missing on Role+Action** (#456) — custom `Action` no longer swaps the item to `handleMenuItem:`. Role+Action keeps the system selector (`orderFrontStandardAboutPanel:` etc.) and overrides via `setTarget:` on the app delegate (Electron/Qt pattern), so AppKit still draws role chrome.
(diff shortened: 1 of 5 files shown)


lkmavi:
## Manual verification (macOS)

Verified on this branch with `examples/menu`:

- [x] Window menu present in menu bar after `SetMenu` (Accessibility: Example App | Tools | Window)
- [x] Window → Close menu action closes the About panel (AXDialog)
- [x] ⌘W (key code 13) closes the About panel while it is focused
- [x] CI green on PR

Ready to merge when you want.

kolkov (dismissed):
Excellent fix, @lkmavi. This matches exactly the enterprise pattern we found in our research (GLFW, Qt6, Flutter, Electron — all auto-preserve the Window menu across menu bar replacement).

**Validated:**

- Retain/Release balanced on the normal path. Detach order is correct: `removeItem:` → `ClearMenuActions` → `removeAllItems` — Window menu callbacks survive the wipe.
- `setWindowsMenu:` re-registration in `restoreWindowsMenu` ensures AppKit resumes auto-populating the window list.
- Fallback path (Window menu stripped before capture) correctly wraps `NSApp.windowsMenu` in a new item via `NewMenuItemWithSubmenu`.
- All new functions nil-safe (`findWindowsMenuItem`, `restoreWindowsMenu`, plus `p.app == nil` guards on `getAppMenu`/`getWindowMenu`).
- No silent error discards.
- CHANGELOG accurate.

**Minor note (non-blocking):** `NewMenuItemWithSubmenu` uses `alloc+init` without `autorelease`, so the fallback-created item has retain count 1 + 1 from `addItem:`. This is a pre-existing pattern (not introduced by this PR) and AppKit manages the lifecycle through the menu bar — no leak in practice.

Approved.

lkmavi:
@kolkov Fixed a minor note, ready for release

